PR #218, opened 2024-12-19
Arc debug firmware multiple features

lgojic-tt:
Features worked on in this pr
 - Reworked communication with arc_dbg_fw
 - Instruction injection for dynamic logging
 - Reseting support for the core where arc_dbg_fw runs
 - Graphing the logs


ihamer-tt (commented):
- [ ] I would split graphing-related code into another PR.
- [ ] Go through all comments and make sure they are not stale.

Issue #242, opened 2025-01-29
Expose reading/writing configuration/debug registers


tt-vjovanovic:
We have `TensixDebug` class that knows how to read/write debug/configuration registers (`read_tensix_register`/`write_tensix_register`).
It would be nice to expose those functions in nicer way through library and also as commands in app.

PR #285, opened 2025-03-12
Added read/write tensix register to lib

adjordjevic-TT:
Issue: https://github.com/tenstorrent/tt-exalens/issues/242

Added functions read_tensix_register and write_tensix_register. There are options to specify register by its name or by TensixRegisterDescription object (or its subclasses).

Added test to test_lib to.

### DIFF
--- a/test/ttexalens/unit_tests/test_lib.py
+++ b/test/ttexalens/unit_tests/test_lib.py
@@ -249,17 +249,23 @@ class TestReadWrite(unittest.TestCase):
         ]
     )
     def test_write_read_tensix_register(self, core_loc, register, value):
+        """Test writing and reading tensix registers"""
 
+        # Storing the original value of the register
         original_value = lib.read_tensix_register(core_loc, register)
 
+        # Writing a value to the register and reading it back
         lib.write_tensix_register(core_loc, register, value)
         ret = lib.read_tensix_register(core_loc, register)
 
+        # Checking if the value was written and read correctly
         self.assertEqual(ret, value)
 
+        # Writing the original value back to the register and reading it
         lib.write_tensix_register(core_loc, register, original_value)
         ret = lib.read_tensix_register(core_loc, register)
 
+        # Checking if read value is equal to the original value
         self.assertEqual(ret, original_value)
 
 
--- a/test/ttexalens/unit_tests/test_lib.py
+++ b/test/ttexalens/unit_tests/test_lib.py
@@ -248,17 +248,19 @@ class TestReadWrite(unittest.TestCase):
             ("0,0", DebugRegisterDescription(address=0x54), 18),
         ]
     )
-    def test_read_write_tensix_register(self, core_loc, register, value):
+    def test_write_read_tensix_register(self, core_loc, register, value):
 
-        lib.write_tensix_register(core_loc, value, register)
+        original_value = lib.read_tensix_register(core_loc, register)
+
+        lib.write_tensix_register(core_loc, register, value)
         ret = lib.read_tensix_register(core_loc, register)
 
-        import time
-
-        print(ret, value)
-        time.sleep(2)
-
         self.assertEqual(ret, value)
+
+        lib.write_tensix_register(core_loc, register, original_value)
+        ret = lib.read_tensix_register(core_loc, register)
+
+        self.assertEqual(ret, original_value)
 
 
 class TestRunElf(unittest.TestCase):
--- a/test/ttexalens/unit_tests/test_lib.py
+++ b/test/ttexalens/unit_tests/test_lib.py
@@ -243,9 +243,16 @@ class TestReadWrite(unittest.TestCase):
 
     @parameterized.expand(
         [
-            ("0,0", ConfigurationRegisterDescription(index=60, mask=0xF, shift=0), 1),
-            ("0,0", ConfigurationRegisterDescription(index=1, mask=0x1E000000, shift=25), 2),
-            ("0,0", DebugRegisterDescription(address=0x54), 18),
+            ("0,0", ConfigurationRegisterDescription(index=60, mask=0xF, shift=0), 1),  # UNPACK_CONFIG0_out_data_format
+            (
+                "0,0",
+                ConfigurationRegisterDescription(index=1, mask=0x1E000000, shift=25),
+                2,
+            ),  # ALU_FORMAT_SPEC_REG2_Dstacc
+            ("0,0", DebugRegisterDescription(address=0x54), 18),  # RISCV_DEBUG_REG_DBG_BUS_CNTL_REG
+            ("0,0", "ALU_FORMAT_SPEC_REG2_Dstacc", 6),
+            ("0,0", "RISCV_DEBUG_REG_DBG_ARRAY_RD_EN", 1),
+            ("0,0", "RISCV_DEBUG_REG_DBG_INSTRN_BUF_CTRL0", 9),
         ]
     )
     def test_write_read_tensix_register(self, core_loc, register, value):
--- a/test/ttexalens/unit_tests/test_lib.py
+++ b/test/ttexalens/unit_tests/test_lib.py
@@ -239,17 +239,24 @@ class TestReadWrite(unittest.TestCase):
         with self.assertRaises((util.TTException, ValueError)):
             lib.write_to_device(core_loc, address, data, device_id)
 
+    from ttexalens.device import ConfigurationRegisterDescription, DebugRegisterDescription
+
     @parameterized.expand(
         [
-            ("0,0", True, 60, 0xF, 0, 1),
-            ("0,0", True, 1, 0x1E000000, 25, 2),
-            ("0,0", False, 0x54, 0x1E000000, 25, 18),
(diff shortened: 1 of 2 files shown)


ttexalens/tt_exalens_lib.py
```diff
@@ -350,3 +350,128 @@ def arc_msg(
...
+    validate_device_id(device_id, context)
+
+    if not isinstance(core_loc, OnChipCoordinate):
+        core_loc = OnChipCoordinate.create(core_loc, device=device)
+
+    if isinstance(register, str):
+        register = device.get_tensix_register_description(register)
+        address = register.address
+    elif isinstance(register, TensixRegisterDescription):
+        address = (
```
tt-vjovanovic, line 447: We don't want to manipulate address given by the user.

ttexalens/tt_exalens_lib.py
```diff
@@ -350,3 +350,128 @@ def arc_msg(
...
+            register.address + device._get_tensix_register_base_address(register)
+            if isinstance(register, ConfigurationRegisterDescription)
+            else register.address
+        )
+    else:
+        TTException(
+            f"Invalid register type. Must be an str or instance of TensixRegisterDescription or its subclasses, but got {type(register)}"
+        )
+
+    if isinstance(register, ConfigurationRegisterDescription):
```
tt-vjovanovic, line 457: Same as for other function. Reuse existing method `TensixDebug.write_tensix_register`

test/ttexalens/unit_tests/test_lib.py
```diff
@@ -239,6 +239,42 @@ def test_invalid_write(self, core_loc, address, data, device_id):
...
+                ConfigurationRegisterDescription(index=1, mask=0x1E000000, shift=25),
+                2,
+            ),  # ALU_FORMAT_SPEC_REG2_Dstacc
+            ("0,0", DebugRegisterDescription(address=0x54), 18),  # RISCV_DEBUG_REG_DBG_BUS_CNTL_REG
+            ("0,0", "ALU_FORMAT_SPEC_REG2_Dstacc", 6),
+            ("0,0", "RISCV_DEBUG_REG_DBG_ARRAY_RD_EN", 1),
+            ("0,0", "RISCV_DEBUG_REG_DBG_INSTRN_BUF_CTRL0", 9),
+        ]
+    )
+    def test_write_read_tensix_register(self, core_loc, register, value):
```
tt-vjovanovic, line 258: You should also add sad path tests. Take a look at `test_invalid_inputs_read`, `test_invalid_write_word` and `test_invalid_write`

ttexalens/tt_exalens_lib.py
adjordjevic-TT: Idea is that user provides index for configuration register so we need to calculate address. When config register is referenced by name (string) function get_tensix_register_description calculates correct address given for given index (from register_map). The safest place to add base address (when user provides object instead of string) seemed to be directly in write function in order not to break something else. We can discuss if we would rather put it somewhere else (get_tensix_register_description or directly into class ConfigurationRegisterDescription).

ttexalens/tt_exalens_lib.py
```diff
@@ -350,3 +350,79 @@ def arc_msg(
...
+    from ttexalens.device import TensixRegisterDescription, ConfigurationRegisterDescription
+    from ttexalens.debug_tensix import TensixDebug
+    from ttexalens.debug_risc import RiscLoader, RiscDebug, RiscLoc
+
+    context = check_context(context)
+    validate_device_id(device_id, context)
+    device = context.devices[device_id]
+
+    if value < 0:
+        raise ValueError("Value must be greater than or equal to 0, but got {value}")
```
tt-vjovanovic, line 419: string format not enabled

ttexalens/tt_exalens_lib.py
```diff
@@ -350,3 +350,79 @@ def arc_msg(
...
+
+    from ttexalens.device import TensixRegisterDescription, ConfigurationRegisterDescription
+    from ttexalens.debug_tensix import TensixDebug
+    from ttexalens.debug_risc import RiscLoader, RiscDebug, RiscLoc
+
+    context = check_context(context)
+    validate_device_id(device_id, context)
+    device = context.devices[device_id]
+
+    if value < 0:
```
tt-vjovanovic, line 418: We should also verify upper limit, but maybe that can be done in debug_tensix where we definitely have `TensixRegisterDescription` instance and register mask, so we can detect max value...

ttexalens/debug_tensix.py
```diff
@@ -176,20 +179,25 @@ def read_tensix_register(self, name: str) -> int:
...
+                register (str | TensixRegisterDescription): Name of the configuration or debug register or instance of ConfigurationRegisterDescription or DebugRegisterDescription.
                 val (int): Value to write
         """
         device = self.context.devices[self.device_id]
-        register = device.get_tensix_register_description(name)
+
+        if isinstance(register, str):
+            register = device.get_tensix_register_description(register)
+        elif isinstance(register, ConfigurationRegisterDescription):
+            register = register.clone(device._get_tensix_register_base_address(register))
```
tt-vjovanovic, line 194: Since now, user can create own instance of `ConfigurationRegisterDescription` we need to verify that index is correct (test range 0 ... max index)...

tt-vjovanovic, line 194: Note that original function does verification that base address exists in device implementation:
```python
            base_address = self._get_tensix_register_base_address(register_description)
            if base_address != None:
                return register_description.clone(base_address)
            else:
                raise ValueError(f"Unknown tensix register base address for register: {register_name}")
```

test/ttexalens/unit_tests/test_lib.py
```diff
@@ -239,6 +239,64 @@ def test_invalid_write(self, core_loc, address, data, device_id):
...
+        self.assertEqual(ret, original_value)
+
+    @parameterized.expand(
+        [
+            ("abcd", ConfigurationRegisterDescription(), 0, 0),  # Invalid core_loc string
+            ("-10", ConfigurationRegisterDescription(), 0, 0),  # Invalid core_loc string
+            ("0,0", ConfigurationRegisterDescription(), 0, -1),  # Invalid device_id
+            ("0,0", ConfigurationRegisterDescription(), 0, 112),  # Invalid device_id (too high)
+            ("0,0", DebugBusSignalDescription(), 0, 0),  # Invalid register type
+            ("0,0", "invalid_register_name", 0, 0),  # Invalid register name
```
tt-vjovanovic, line 287: Maybe these as well:
```python
("0,0", 0xFFB12345, 0, 0),  # Address alone is not enough to represent index
("0,0", ConfigurationRegisterDescription(index=-1, mask=0xF, shift=0), 1),  # index cannot be negative
("0,0", ConfigurationRegisterDescription(index=0xFFFFFFFF, mask=0xF, shift=0), 1),  # index cannot be that large
```

ttexalens/tt_exalens_lib.py
```diff
@@ -350,3 +350,79 @@ def arc_msg(
...
+    Args:
+            core_loc (str | OnChipCoordinate): Either X-Y (noc0/translated) or X,Y (logical) location of a core in string format, dram channel (e.g. ch3), or OnChipCoordinate object.
+            register (str | TensixRegisterDescription): Tensix register to read from (name or class instance).
+            device_id (int, default 0):	ID number of device to read from.
+            context (Context, optional): TTExaLens context object used for interaction with device. If None, global context is used and potentailly initialized.
+
+    Returns:
+            int: Value of the configuration or debug register specified.
+    """
+    from ttexalens.device import TensixRegisterDescription, ConfigurationRegisterDescription
```
tt-vjovanovic: `ConfigurationRegisterDescription` not used in this method

ttexalens/tt_exalens_lib.py
```diff
@@ -350,3 +350,79 @@ def arc_msg(
...
+
+    Args:
+            core_loc (str | OnChipCoordinate): Either X-Y (noc0/translated) or X,Y (logical) location of a core in string format, dram channel (e.g. ch3), or OnChipCoordinate object.
+            register (str | TensixRegisterDescription): Tensix register to read from (name or class instance).
+            value (int): Value to write to the register.
+            device_id (int, default 0):	ID number of device to read from.
+            context (Context, optional): TTExaLens context object used for interaction with device. If None, global context is used and potentailly initialized.
+    """
+
+    from ttexalens.device import TensixRegisterDescription, ConfigurationRegisterDescription
```
tt-vjovanovic: `ConfigurationRegisterDescription` not used in this method

ttexalens/tt_exalens_lib.py
```diff
@@ -350,3 +350,79 @@ def arc_msg(
...
+    from ttexalens.debug_tensix import TensixDebug
+    from ttexalens.debug_risc import RiscLoader, RiscDebug, RiscLoc
```
tt-vjovanovic: `RiscLoader, RiscDebug, RiscLoc` not used in this method

tt-vjovanovic (approved):

PR #286, opened 2025-03-14
Update unit tests and blackhole device implementation

miacim:
- Comment out TRISC2 in test_callstack.py due to a known bug.
- Add method to check for blackhole devices in test_lib.py.
- Use debug bus to check PC register in test_risc_debug.py.
- Enhance blackhole device with debug bus signal descriptions.

### DIFF
--- a/test/ttexalens/unit_tests/test_risc_debug.py
+++ b/test/ttexalens/unit_tests/test_risc_debug.py
@@ -116,6 +116,13 @@ class TestDebugging(unittest.TestCase):
                 self.program_base_address + expected,
                 f"Pc should be {expected} + program_base_addres ({self.program_base_address + expected}).",
             )
+        else:
+            self.assertEqual(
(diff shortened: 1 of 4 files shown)


test/ttexalens/unit_tests/test_risc_debug.py
```diff
@@ -109,18 +109,12 @@ def write_program(self, addr, data):
 
     def assertPcEquals(self, expected):
         """Assert PC register equals to expected value."""
-        self.assertEqual(
-            self.rdbg.read_gpr(self.pc_register_index),
```
tt-vjovanovic, line 113: Should this be else part?

tt-vjovanovic (approved):

test/ttexalens/unit_tests/test_risc_debug.py
miacim: Yes, you are correct. We will remove it when we disable grayskull tests.

PR #289, opened 2025-03-17
Adding read/write tensix register as a command

adjordjevic-TT:
Issue: https://github.com/tenstorrent/tt-exalens/issues/242

Added two commands (dump-reg and write-reg) for reading and writing to configuration and debug registers in tensix cores. This is done through read/write_tensix_register methods from TensixDebug.

Commands are added to wheel tests.


ttexalens/cli_commands/dump-config-reg.py
```diff
@@ -8,7 +8,7 @@
 Options:
   <config-reg>  Configuration register name to dump. Options: [all, alu, pack, unpack] Default: all
   -d <device>   Device ID. Optional. Default: current device
-  -l <loc>      Core location in X-Y or R,C format
+  -l <loc>      Core location in X-Y or R,C format. Defualt: current core
```
tt-vjovanovic, line 11: Nit, typo: `Default`

ttexalens/cli_commands/dump-config-reg.py
```diff
@@ -88,6 +89,9 @@ def run(cmd_text, context, ui_state: UIState = None):
             debug_tensix = TensixDebug(loc, device.id(), context)
             device = debug_tensix.context.devices[debug_tensix.device_id]
 
+            if cfg not in possible_registers:
```
tt-vjovanovic, line 92: Consider doing this outside of the for loops...

ttexalens/cli_commands/tensix-reg.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def convert_to_int(param: str) -> int:
+    try:
+        if param.startswith("0x"):
+            return int(param, 16)
+        else:
+            return int(param)
+    except ValueError:
+        raise ValueError("Invalid parameter. Expected an integer.")
```
tt-vjovanovic, line 63: Potential user problem, `convert_value` method knows how to print bools, enums, etc. It is OK not to parse everything, but just saying that users might have problem with it...

Also, maybe update message:
```python
raise ValueError(f"Invalid write value '{param}'. Expected an integer.")
```

ttexalens/cli_commands/tensix-reg.py
```diff
@@ -0,0 +1,124 @@
...
+            index=reg_params[0], mask=reg_params[1], shift=reg_params[2], data_type=DATA_TYPE[data_type]
+        )
+    elif reg_type == "dbg":
+        if len(reg_params) != 1:
+            raise TTException(
+                "Invalid number of parameters for debug register since it requires 1 parameter (address)."
+            )
+        return DebugRegisterDescription(address=reg_params[0], data_type=DATA_TYPE[data_type])
+    else:
+        raise ValueError(f"Unknown register type: {reg_type}. Possible values: {reg_types}")
```
tt-vjovanovic, line 97: Why can't users use names that we have defined, or that Metal/LLK uses? You can always search for register name in device class and print it is not available if that core doesn't have that register (in our "database")...
Example: `NOC_NODE_ID` or `TRISC_RESET_PC_SEC0_PC`

ttexalens/cli_commands/tensix-reg.py
```diff
@@ -0,0 +1,124 @@
...
+  --type <data-type>  Data type of the register. Options: [INT_VALUE, ADDRESS, MASK, FLAGS, TENSIX_DATA_FORMAT]. Default: INT_VALUE
+  --write <value>     Value to write to the register. If not given, register is dumped instead.
+  -d <device>         Device ID. Optional. Default: current device
+  -l <loc>            Core location in X-Y or R,C format. Default: current core
+
+Description:
+  Prints/writes to the specified register, at the specified location and device.
+
+Examples:
+  reg cfg(60,0xf,0)               # Prints configuration register with index 60, mask 0xf, shift 0
```
tt-vjovanovic, line 23: Not sure what this does on blackhole...
On Wormhole it is targeting `UNPACK_CONFIG0_out_data_format`, but we don't have register stored for blackhole with these settings.

ttexalens/cli_commands/tensix-reg.py
adjordjevic-TT: Added option to use name instead of giving configuration. Since there are a lot of register names is it sensible to add option to list all names?

adjordjevic-TT: Maybe the option would be to add suggestion tab as user writes register name.

ttexalens/cli_commands/tensix-reg.py
adjordjevic-TT: Added options to write boolean list (False,True...) or TensixDataFormat by giving name (ex. Float16_b)

ttexalens/cli_commands/tensix-reg.py
```diff
@@ -0,0 +1,163 @@
...
+
+def convert_reg_params(reg_params: str) -> list[int]:
+    params = []
+    for param in reg_params.split(","):
+        params.append(convert_to_int(param))
+
+    return params
+
+
+def convert_write_value(value: str) -> int:
```
tt-vjovanovic, line 77: Note that this is hard to maintain if it is not next to `convert_value` method in utils. Maybe update name if you move to utils..

tt-vjovanovic (approved):

ttexalens/cli_commands/tensix-reg.py
adjordjevic-TT: Moved to util. Changed convert_value to convert_int_to_data_type and changed convert_write_value to convert_data_type_to_int.

ttexalens/cli_commands/tensix-reg.py
adjordjevic-TT: Added search feature using wildcard pattern. You can also define max-regs (default 10), negative values represent all.

Issue #293, opened 2025-03-20
RISCV Core Lockup During Halt/Resume Operations


miacim:
## Description

Repeatedly halting and resuming a RISCV core while load instructions are in-flight can trigger a core lockup. Specifically, halting (with COMMAND_DEBUG_MODE | COMMAND_HALT) repurposes the memory port, discarding in-flight memory read responses. Discarding these responses corrupts load results, causing instructions to retire with incorrect data and rapidly leading to a complete core lockup.

### Test 1

Run this test loop on the RISCV core:

```asm
li t3, 0
b_loop:
addi t3, t3, 1 # Counter increment
lw t1, 0(x0) # L1 read
lw t2, 0(x0) # L1 read
j b_loop
```
Perform multiple halt/resume cycles (50+ iterations):
Issue COMMAND_DEBUG_MODE | COMMAND_HALT
Issue COMMAND_CONTINUE
Check whether core locks up.

### Test 2
Run on brisc

```asm
li t0, 0xFFEC1000 # T0 -> B mailbox (when reading on RISCV B)
li t3, 0
b_loop:
addi t3, t3, 1 # Counter increment
lw t1, 0(t0) # Mailbox pop
j b_loop
```

Run on trisc0

```asm
li t0, 0xFFEC0000 # T0 -> B mailbox (when writing on RISCV T0)
li t3, 0
t0_loop:
addi t3, t3, 1 # Counter increment
slli t1, t3, 20 # Play with the 20 to vary how often mailbox writes happen
bne t1, x0, t0_loop
sw t3, 0(t0) # Mailbox push
j t0_loop
```

Attempt to halt brisc and then resume it. Repeat this halt/resume number of times. Verify that core lock up.

** Check if issue is reproducible on all generations (wormhole, blackhole, ...) **

## Partial Mitigation

Issuing debug mode first, followed by a short delay, then issuing debug mode | halt can mitigate the issue in standard load scenarios.

Validate that this will fix Test1.

Update documentation to clearly warn users about this issue and its mitigation.

miacim:
Test need to be run on Quasar.

miacim:
It turns out that issue was not reproducible on blackhole, only wormhole issue.

PR #306, opened 2025-03-31
Exposing lockup issue when halting cores

miacim:
Adding tests that exposes potential issues with halting cores. There is scenario explained in bug #293, that can cause core lockup.

### DIFF
--- a/test/ttexalens/unit_tests/core_simulator.py
+++ b/test/ttexalens/unit_tests/core_simulator.py
@@ -7,6 +7,7 @@ from ttexalens.context import Context
 from ttexalens.debug_risc import RiscLoader, RiscDebug, RiscLoc, get_register_index, get_risc_id
 import time
 
+
 class RiscvCoreSimulator:
     """Class to simulate and control a RISC-V core for testing purposes."""
 
@@ -141,13 +142,13 @@ class RiscvCoreSimulator:
 
     def read_gpr(self, reg_num: int) -> int:
         """Read general purpose register value.
-        
+
         Args:
             reg_num: Register number (0-31)
-            
+
         Returns:
             int: Register value
-            
+
         Raises:
             ValueError: If register number is invalid
         """
@@ -157,10 +158,10 @@ class RiscvCoreSimulator:
 
     def get_reg_num(self, reg_name: str) -> int:
         """Get register number from name.
-        
+
         Args:
             reg_name: Register name (e.g. "zero", "ra", "sp", "t0", etc.)
-            
+
         Returns:
             int: Register number
         """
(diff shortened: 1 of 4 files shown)


test/ttexalens/unit_tests/core_simulator.py
```diff
@@ -0,0 +1,168 @@
...
+        self.write_data_checked(self.program_base_address + addr, data)
+
+    def write_data_checked(self, addr: int, data: int):
+        """Write data to memory and verify it was written correctly."""
+        lib.write_words_to_device(self.core_loc, addr, data, context=self.context)
+        assert self.read_data(addr) == data, f"Data verification failed at address {addr:x}"
+
+    def read_data(self, addr: int) -> int:
+        """Read data from memory at specified address."""
+        ret = lib.read_words_from_device(self.core_loc, addr, context=self.context)
```
tt-vjovanovic, line 72: We can switch to "new" API from months ago :)
```python
return lib.read_word_from_device(self.core_loc, addr, context=self.context)
```

test/ttexalens/unit_tests/core_simulator.py
```diff
@@ -0,0 +1,168 @@
...
+        Args:
+            reg_num: Register number (0-31)
+
+        Returns:
+            int: Register value
+
+        Raises:
+            ValueError: If register number is invalid
+        """
+        if not 0 <= reg_num <= 31:
```
tt-vjovanovic, line 155: Isn't PC with index 32?

ttexalens/debug_risc.py
```diff
@@ -587,10 +599,8 @@ def ensure_reading_configuration_register(self):
                         yield brisc_debug
                 finally:
                     # Return BRISC in reset
-                    self.risc_debug.set_reset_signal(1)
-                    assert (
-                        self.risc_debug.is_in_reset()
-                    ), f"RISC at location {self.risc_debug.location} is not in reset."
+                    brisc_debug.set_reset_signal(1)
```
tt-vjovanovic, line 602: Please save 4 bytes from 0 (where brisc_loader will overwrite with infinite loop) and restore them here in finally...

test/ttexalens/unit_tests/test_risc_debug.py
```diff
@@ -385,6 +385,37 @@ def test_continue(self):
         # Verify value at address
         self.assertEqual(self.read_data(addr), 0x87654000)
 
+    def test_core_lockup(self):
```
tt-vjovanovic, line 388: Not comment for this test, but should we reuse `RiscvCoreSimulator` in test class?

tt-vjovanovic (approved):

test/ttexalens/unit_tests/test_risc_debug.py
miacim: when we start with refactoring tests, definitely we should reuse it.

Issue #292, opened 2025-03-19
[Feature Request] Support backtracing kernels compiled by tt-metal


jbaumanTT:
Attempting to backtrace a tt-metal kernel, I get this error:

bt /home/jbauman/.cache/tt-metal-cache/e1435bdc91/1052672/kernels/bmm_large_block_zm_fused_bias_activation/2981475175402633130/trisc2/trisc2.elf -r 3
Reading from RiscV debug registers failed (debug read valid bit is set to 0). Run `srs 0` to check if core is active.
Reading from RiscV debug registers failed (debug read valid bit is set to 0). Run `srs 0` to check if core is active.
Reading from RiscV debug registers failed (debug read valid bit is set to 0). Run `srs 0` to check if core is active.
Reading from RiscV debug registers failed (debug read valid bit is set to 0). Run `srs 0` to check if core is active.
Reading from RiscV debug registers failed (debug read valid bit is set to 0). Run `srs 0` to check if core is active.
------------------------------------------------------------------------------  -------------------------  ------------------------------------------------------------------------------------
python_env/lib/python3.10/site-packages/ttlens/cli.py:361                       main_loop                  new_navigation_suggestions = found_command["module"].run(cmd_raw, context, ui_state)
python_env/lib/python3.10/site-packages/ttlens/ttlens_commands/callstack.py:60  run                          callstack = loader.get_callstack(elf_path, limit, stop_on_main)
python_env/lib/python3.10/site-packages/ttlens/debug_risc.py:951                get_callstack                  elf = read_elf(self.context.server_ifc, elf_path)
python_env/lib/python3.10/site-packages/ttlens/parse_elf.py:725                 read_elf                         dwarf = elf.get_dwarf_info()
python_env/lib/python3.10/site-packages/elftools/elf/elffile.py:275             get_dwarf_info                     dwarf_section = self._read_dwarf_section(
python_env/lib/python3.10/site-packages/elftools/elf/elffile.py:801             _read_dwarf_section                  reloc_handler.apply_section_relocations(
python_env/lib/python3.10/site-packages/elftools/elf/relocation.py:238          apply_section_relocations              self._do_apply_relocation(stream, reloc, symtab)
python_env/lib/python3.10/site-packages/elftools/elf/relocation.py:293          _do_apply_relocation                     raise ELFRelocationError(
[...]

PR #313, opened 2025-04-09
Added callstack (backtrace) support for metal kernels

adjordjevic-TT:
Issue: https://github.com/tenstorrent/tt-exalens/issues/292

Now command takes as arguments list of elf paths and offsets. Offsets represent location in memory where kernel is loaded (for firmware it should be 0 or can be left empty), for kernels it should be kernel_address variable from firmware code.

This change also relies on pyelftools patch regarding relocation support for RISC-V.

Currently, it is not possible to load kernel directly from lens so we rely on metal for that. Intended use is you run metal code, it hangs you go to lens and use callstack command providing list of elf_paths for kernels that are used and their offsets (kernel addresses).

### DIFF
--- a/ttexalens/hw/tensix/blackhole/blackhole.py
+++ b/ttexalens/hw/tensix/blackhole/blackhole.py
@@ -673,6 +673,3167 @@ class BlackholeDevice(Device):
         "trisc1_pc": DebugBusSignalDescription(rd_sel=1, daisy_sel=7, sig_sel=2 * 7 + 1, mask=0x3FFFFFFF),
         "trisc2_pc": DebugBusSignalDescription(rd_sel=1, daisy_sel=7, sig_sel=2 * 8 + 1, mask=0x3FFFFFFF),
         "ncrisc_pc": DebugBusSignalDescription(rd_sel=1, daisy_sel=7, sig_sel=2 * 12 + 1, mask=0x3FFFFFFF),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_IBUFFER_STALL": DebugBusSignalDescription(
+            rd_sel=0, daisy_sel=1, sig_sel=13, mask=0x80
+        ),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_RISC_CFG_STALL": DebugBusSignalDescription(
+            rd_sel=0, daisy_sel=1, sig_sel=13, mask=0x40
+        ),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_RISC_GPR_STALL": DebugBusSignalDescription(
+            rd_sel=0, daisy_sel=1, sig_sel=13, mask=0x20
+        ),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_RISC_TDMA_STALL": DebugBusSignalDescription(
+            rd_sel=0, daisy_sel=1, sig_sel=13, mask=0x10
+        ),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_PREV_GEN_NO__3_0": DebugBusSignalDescription(
+            rd_sel=3, daisy_sel=1, sig_sel=12, mask=0xF0000000
+        ),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_PREV_GEN_NO__7_4": DebugBusSignalDescription(
+            rd_sel=0, daisy_sel=1, sig_sel=13, mask=0xF
+        ),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_LSQ_HEAD_VALID": DebugBusSignalDescription(
+            rd_sel=3, daisy_sel=1, sig_sel=12, mask=0x8000000
+        ),
+        "DEBUG_BUS_THREAD_STATE[0]_I_TTSYNC_DBG_BITS_LSQ_HEAD_RSRCS_WR_TDMA": DebugBusSignalDescription(
+            rd_sel=3, daisy_sel=1, sig_sel=12, mask=0x4000000
+        ),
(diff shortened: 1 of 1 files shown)


ttexalens/debug_risc.py
```diff
@@ -954,26 +954,73 @@ def run_elf(self, elf_path: str):
             not self.risc_debug.is_halted() or self.risc_debug.read_status().is_ebreak_hit
         ), f"RISC at location {self.risc_debug.location} is still halted, but not because of ebreak."
 
-    def get_callstack(self, elf_path: str, limit: int = 100, stop_on_main: bool = True):
+    def _read_elfs(self, elf_paths: List[str], offsets: List[int]) -> List[dict]:
+        if not isinstance(elf_paths, list):
+            elf_paths = [elf_paths]
+        offsets = [None] * len(elf_paths) if offsets is None else offsets
```
tt-vjovanovic, line 960: Where do we verify lists are of the same size? What is the strategy if they aren’t?

adjordjevic-TT: Added check if they are the same. If not prints message and exit.

ttexalens/debug_risc.py
```diff
@@ -954,26 +954,73 @@ def run_elf(self, elf_path: str):
...
+    ):
         callstack = []
         with self.risc_debug.ensure_halted():
-            elf = read_elf(self.context.server_ifc, elf_path)
+
+            # Reading the elf files for given paths and offsets
+            elfs = self._read_elfs(elf_paths, offsets)
+
+            # Choosing first elf as starting point
+            elf = elfs[0]
```
tt-vjovanovic, line 979: Make a small method that will search elf's to find which elf is referencing current PC. Method should return FrameDescription and elf, so later we can get file, line, function description...

ttexalens/debug_risc.py
```diff
@@ -954,26 +954,73 @@ def run_elf(self, elf_path: str):
...
+                    return []
+
             frame_pointer = frame_description.read_previous_cfa()
             i = 0
             while i < limit:
                 file_line = elf["dwarf"].find_file_line_by_address(pc)
                 function_die = elf["dwarf"].find_function_by_address(pc)
-                if function_die is not None and function_die.category == "inlined_function":
+
+                # Skipping lexical blocks since we do not print them
```
tt-vjovanovic, line 1005: Is this happening for subprogram as well or just for inlined functions? If it is inlined functions specific, maybe you should deal with it in that if branch :)

ttexalens/parse_elf.py
```diff
@@ -329,13 +336,37 @@ def resolved_type(self):
         Resolve to underlying type
         TODO: test typedefs, this looks overly complicated
         """
-        if self.tag == "DW_TAG_typedef":
+        if self.tag == "DW_TAG_typedef" and self.local_offset != None:
             typedef_DIE = self.cu.find_DIE_at_local_offset(self.local_offset)
+            dwarf_die = self.dwarf_die.get_DIE_from_attribute("DW_AT_type")
```
tt-vjovanovic, line 341: Remove test code

ttexalens/parse_elf.py
```diff
@@ -329,13 +336,37 @@ def resolved_type(self):
...
+        elif self.tag == "DW_TAG_const_type" and self.local_offset != None:
+            typedef_DIE = self.cu.find_DIE_at_local_offset(self.local_offset)
+            dwarf_die = self.dwarf_die.get_DIE_from_attribute("DW_AT_type")
+            if typedef_DIE.dwarf_die != dwarf_die:
+                print("ERROR")
+            if typedef_DIE:  # If typedef, recursivelly do it
+                return typedef_DIE.resolved_type
+        elif self.tag == "DW_TAG_volatile_type" and self.local_offset != None:
+            typedef_DIE = self.cu.find_DIE_at_local_offset(self.local_offset)
+            dwarf_die = self.dwarf_die.get_DIE_from_attribute("DW_AT_type")
```
tt-vjovanovic, line 355: Remove test code

ttexalens/parse_elf.py
```diff
@@ -329,13 +336,37 @@ def resolved_type(self):
...
+            if typedef_DIE.dwarf_die != dwarf_die:
+                print("ERROR")
+            if typedef_DIE:  # If typedef, recursivelly do it
+                return typedef_DIE.resolved_type
+        elif self.category != "type" and "DW_AT_type" in self.attributes and self.local_offset != None:
             my_type_die = self.cu.find_DIE_at_local_offset(self.local_offset)
-            if my_type_die.tag == "DW_TAG_typedef":
+            dwarf_die = self.dwarf_die.get_DIE_from_attribute("DW_AT_type")
```
tt-vjovanovic, line 362: Remove test code

ttexalens/cli_commands/callstack.py
```diff
@@ -3,12 +3,16 @@
 # SPDX-License-Identifier: Apache-2.0
 """
 Usage:
-  callstack <elf-file> [-r <risc>] [-m <max-depth>] [-v] [-d <device>] [-l <loc>]
+  callstack <elf-files> [-o <offsets>] [-r <risc>] [-m <max-depth>] [-v] [-d <device>] [-l <loc>]
```
tt-vjovanovic, line 6: If I remember correctly we should have library function for getting callstack as well... See if that is true and update API as well...

adjordjevic-TT: It appears we don't have lib function for callstack. Should I add it?

tt-vjovanovic: If we want them to use this in scripting, we should definitely add a method in lib for callstack

ttexalens/debug_risc.py
adjordjevic-TT: As I understand it is inlined function specific since it is the only one we are unwinding. The problem arises when lexical block is last in the chain (so we hit it first) so I skipped them until we hit inlined function. Other option is to add "or function_die.category == "lexical_block" to inlined function branch and then skip it inside that branch which may make more sense.

ttexalens/cli_commands/callstack.py
miacim: Command should start using API.

miacim (approved):

Issue #317, opened 2025-04-14
Blackhole Debug Bus Signals


miacim:
Extend debug bus signals for blackhole explored in [dbus-spec](https://github.com/tenstorrent/tt-llk/tree/lpremovic/dbus-spec).

PR #318, opened 2025-04-14
Adding tensix debug bus signals for blackhole

miacim:
#317 Added list of debug bus signals for blackhole.


lpremovicTT (approved):

miacim:
> These signals are all for tensix cores. Should we explore signals for other coordinates? Like find eth subset, or see if we have something on dram or arc, etc...

Created issue that will address debug bus signals related to eth cores.

Issue #249, opened 2025-02-05
Expose API for printing NOC Registers


miacim:
- Define NOC registers
- Reusable API for fetching and printing registers

PR #324, opened 2025-04-15
Add support for displaying all and specific NOC registers

miacim:
#249 

- Added 'noc all' and 'noc register <reg-names>' CLI commands
- Implemented retrieval and display of all NOC registers
- Added validation and error reporting for unknown register names
- Refactored status display logic to differentiate status from all registers
- Extended Device class with get_noc_register_names() for NOC register filtering

### DIFF
--- a/ttexalens/cli_commands/noc.py
+++ b/ttexalens/cli_commands/noc.py
@@ -4,12 +4,16 @@
 """
 Usage:
     noc status [-d <device>] [--noc <noc-id>] [-l <loc>] [-s]
-    noc dump [-d <device>] [-l <loc>] [-s]
+    noc all [-d <device>] [-l <loc>] [-s]
+    noc register <reg-names> [-d <device>] [--noc <noc-id>] [-l <loc>] [-s]
+
 
 Arguments:
     device-id         ID of the device [default: current active]
     noc-id            Identifier for the NOC (e.g. 0, 1) [default: both noc0 and noc1]
     loc               Location identifier (e.g. 0-0) [default: current active]
+    reg-names         Name of specific NOC register(s) to display, can be comma-separated
+
 
 Options:
     -s, --simple     Print simple output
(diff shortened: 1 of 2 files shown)


ttexalens/cli_commands/noc.py
```diff
@@ -4,23 +4,32 @@
...
+    noc all [-d <device>] [-l <loc>] [-s]
+    noc register <reg-names> [-d <device>] [--noc <noc-id>] [-l <loc>] [-s]
+
 
 Arguments:
     device-id         ID of the device [default: current active]
     noc-id            Identifier for the NOC (e.g. 0, 1) [default: both noc0 and noc1]
     loc               Location identifier (e.g. 0-0) [default: current active]
+    reg-names         Name of specific NOC register(s) to display, can be comma-separated
+
```
adjordjevic-TT, line 16: Maybe we can add search option like we have in tensix-reg so user can search register names using wildcard format (something like noc --search *status* would return names of all status registers).

ttexalens/cli_commands/noc.py
```diff
@@ -177,13 +275,13 @@ def run(cmd_text: str, context: Dict, ui_state: Optional[UIState] = None) -> Lis
     # Parse and validate NOC ID
     if dopt.args["--noc"]:
         try:
-            noc_id = int(dopt.args["--noc"])
+            noc_id = int(dopt.args["<noc-id>"])
             if noc_id not in [0, 1]:
```
adjordjevic-TT, line 279: For consistency maybe change [0,1] to (0,1)

adjordjevic-TT (approved):

tt-vjovanovic (approved):

ttexalens/cli_commands/noc.py
miacim: Make sense. Created issues #328

ttexalens/cli_commands/noc.py
miacim: Yes, that is more performant option also.

PR #325, opened 2025-04-16
Removed job that uses ubuntu-20.04

adjordjevic-TT:
Since ubuntu-20.04 is now retired, job that uses it is removed.

### DIFF
--- a/.github/workflows/release-latest.yml
+++ b/.github/workflows/release-latest.yml
@@ -12,7 +12,6 @@ jobs:
     strategy:
       matrix:
         build-info: [
-          {py-version: "3.8", os: "ubuntu-20.04"},
           {py-version: "3.10", os: "ubuntu-22.04"},
         ]
     name: "Latest Release"


.github/workflows/release-latest.yml
```diff
@@ -12,7 +12,6 @@ jobs:
     strategy:
       matrix:
         build-info: [
-          {py-version: "3.8", os: "ubuntu-20.04"},
```
tt-vjovanovic, line 15: While you are at it, can you change to start using our docker image for release as well instead of using plain one and installing what’s necessary for it?

tt-vjovanovic, line 15: Do we want to add support for 24.04?

tt-vjovanovic:
It is not related to this change, but while you are at it, we should start using metal image as base image for our docker :)

tt-vjovanovic (approved):

.github/workflows/release-latest.yml
miacim: Yes. @adjordjevic-TT  can you create issue to add support for ubuntu 24.04?

miacim (approved):

.github/workflows/release-latest.yml
adjordjevic-TT: Will do in another pr

PR #336, opened 2025-04-17
Remove fallback_tlbs

broskoTT:
These are to be internally defined and used. 
Related to: https://github.com/tenstorrent/tt-umd/issues/737
Blocking further removal from API: https://github.com/tenstorrent/tt-umd/pull/764

### DIFF
--- a/ttexalens/server/lib/src/umd_implementation.cpp
+++ b/ttexalens/server/lib/src/umd_implementation.cpp
@@ -9,11 +9,6 @@
 #include "ttexalensserver/read_tile.hpp"
 #include "umd/device/cluster.h"
 
-static std::string REG_TLB_STR = "REG_TLB";
-static std::string SMALL_READ_WRITE_TLB_STR = "SMALL_READ_WRITE_TLB";
-static std::string LARGE_READ_TLB_STR = "LARGE_READ_TLB";
-static std::string LARGE_WRITE_TLB_STR = "LARGE_WRITE_TLB";
-
 namespace tt::exalens {
 
 umd_implementation::umd_implementation(tt_device* device) : device(device) {}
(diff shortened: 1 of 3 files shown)


ttexalens/server/lib/src/umd_implementation.cpp
```diff
@@ -23,15 +18,15 @@ std::optional<uint32_t> umd_implementation::pci_read32(uint8_t chip_id, uint8_t
...
+    device->read_from_device_reg(&result, chip_id, target, address, sizeof(result));
     return result;
 }
 
 std::optional<uint32_t> umd_implementation::pci_write32(uint8_t chip_id, uint8_t noc_x, uint8_t noc_y, uint64_t address,
                                                         uint32_t data) {
     tt::umd::CoreCoord target = device->get_soc_descriptor(chip_id).get_coord_at({noc_x, noc_y}, CoordSystem::VIRTUAL);
 
-    device->write_to_device(&data, sizeof(data), chip_id, target, address, LARGE_WRITE_TLB_STR);
+    device->write_to_device(&data, sizeof(data), chip_id, target, address);
```
tt-vjovanovic, line 29: Quick question. I wrote that code based on one that was living in budabackend - I just resolved all boolean arguments into tlbs. As we know reg tlb behaves in a special way. So my question here is, do we want to switch to reg tlb? I've seen that people continue to use exalens even if they reset device (they don't restart the exalens app). Will writes still work?

tt-vjovanovic (approved):

ttexalens/server/lib/src/umd_implementation.cpp
broskoTT: The difference is that the TLB is setup in a different way, and I think it uses UC rather than WC memory. But the writes/reads should go in both cases.

I also though I should change this specific line to write_to_device_reg, looking at the code right above this

tt-vjovanovic (approved):

PR #338, opened 2025-04-23
Removed dependency on UMD regarding gtest (fixed failing CI)

adjordjevic-TT:
We previously fetched gtest from the UMD setup. Changes to their setup caused our CI to fail.
Now, we fetch gtest independently, removing this dependency.

Added CMakeLists.txt to test folder (it also makes sense to add it to third_party instead, waiting for Wilder to respond) that fetches gtest using CPM.

Also, copied CPM.cmake from UMD so we don't depend on them for that either.


CMakeLists.txt
```diff
@@ -69,5 +69,6 @@ include(${PROJECT_SOURCE_DIR}/cmake/ttexalens_private.cmake)
 # Add subdirectories
 add_subdirectory(${PROJECT_SOURCE_DIR}/ttexalens)
 add_subdirectory(${PROJECT_SOURCE_DIR}/utils)
+add_subdirectory(${PROJECT_SOURCE_DIR}/test)
```
tt-vjovanovic, line 72: You should move two following line into the new file..

cmake/CPM.cmake
```diff
@@ -0,0 +1,23 @@
+# SPDX-License-Identifier: MIT
```
tt-vjovanovic, line 1: Having a copy of this file makes sense only if we are going to use UMD as CPM reference... Otherwise, we should just reference existing file from UMD submodule... No need to maintain that file...

test/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# Add subdirectories
+add_subdirectory(${PROJECT_SOURCE_DIR}/test/ttexalens/pybind)
```
tt-vjovanovic, line 2: Not sure if this needs to be after adding gtest...

adjordjevic-TT: It works either way, but I agree it makes more sense to place it after adding gtest so I moved it there.

blozano-tt (approved):
LGTM!

PR #342, opened 2025-04-27
Remove tt_device from UMD

broskoTT:
Related to https://github.com/tenstorrent/tt-umd/issues/250

## Changes
- Change all tt_device to tt::umd::Cluster
- Change Cluster constructor to use the new one with ClusterOptions struct
- Simulation is created through Cluster now as well.


ttexalens/server/lib/inc/ttexalensserver/read_tile.hpp
```diff
@@ -46,7 +46,7 @@ enum class TileDataFormat : uint8_t {
 TileDataFormat to_data_format(uint8_t i);
 
 std::optional<std::string> read_tile_implementation(uint8_t chip_id, uint8_t noc_x, uint8_t noc_y, uint64_t address,
-                                                    uint32_t size, uint8_t data_format, tt_device* device);
+                                                    uint32_t size, uint8_t data_format, tt::umd::Cluster* device);
```
tt-vjovanovic, line 49: Maybe we should rename device variable to cluster as well?

tt-vjovanovic: Other places too...

tt-vjovanovic (approved):

ttexalens/server/lib/inc/ttexalensserver/read_tile.hpp
broskoTT: renamed at many places

PR #343, opened 2025-04-28
Fixed a bug in rw tensix register

adjordjevic-TT:
Fixed a bug when getting register description from its name in tensix-reg command. The bug was that base address was not added resulting in reading from or writing to incorrect memory location.

Also added nicer print for TensixRegisterDescription and its subclasses.


ttexalens/cli_commands/tensix-reg.py
```diff
@@ -177,8 +177,7 @@ def run(cmd_text, context, ui_state: UIState = None):
             debug_tensix = TensixDebug(loc, device.id(), context)
 
             if isinstance(register_ref, str):
-                device = debug_tensix.context.devices[debug_tensix.device_id]
-                register = device._get_tensix_register_description(register_ref)
+                register = device.get_tensix_register_description(register_ref)
```
miacim, line 180: Can you create task to use API call in this cli?

miacim (approved):

PR #347, opened 2025-05-03
Initial version of mypy

tt-vjovanovic:
Adding mypy to validate python code.
Configuration file `mypy.ini` has disabled lots of checks.
We should follow up on configuration that is currently set.
After Quasar testing is checked-in, we should remove:
```
disable_error_code = abstract
```


ttexalens/debug_tensix.py
```diff
@@ -259,7 +264,7 @@ def write_tensix_register(self, register: Union[str, TensixRegisterDescription],
     def read_regfile_data(
         self,
         regfile: Union[int, str, REGFILE],
```
adjordjevic-TT, line 266: Maybe we should change all Unions with | since we the earliest python version we support is 3.10.

ttexalens/tt_exalens_lib.py
```diff
@@ -413,23 +427,20 @@ def write_tensix_register(
...
 
-    TensixDebug(core_loc, device_id, context).write_tensix_register(register, value)
+    TensixDebug(coordinate, device_id, context).write_tensix_register(register, value)
 
 
 def callstack(
     core_loc: Union[str, OnChipCoordinate],
     elf_paths: Union[List[str], str],
-    offsets: List[int] = None,
+    offsets: int | List[int | None] = None,
```
adjordjevic-TT, line 443: Similarly to Union, since python 3.10 it is possible to use list, dict, set etc. instead of same things from typing module. Maybe we should create an issue to change this and remove the need for importing things from typing module.

adjordjevic-TT (approved):

miacim (approved):

ttexalens/debug_tensix.py
tt-vjovanovic: Created #361.

ttexalens/tt_exalens_lib.py
tt-vjovanovic: Created #361.

PR #353, opened 2025-05-06
Command that check for mismatches in number of noc transactions between variables and registers

adjordjevic-TT:
Issue: https://github.com/tenstorrent/tt-exalens/issues/231

Added script check_noc_status.py in ttexalens/scripts that checks for mismatches in number of NOC transactions stored in global variables from risc firmware and NOC status registers.

Command first parses firmware elf (only for brisc due to lack of debug hardware in ncrisc) to find addresses of global variables in risc private memory. Then read_riscv_memory from lib is used to read their values. Read_tensix_register from lib is used to read values from corresponding noc status registers. These values are compared and any mismatches are reported. This is done for all devices and core locations.


adjordjevic-TT:
Maybe we will add this functionality as a script rather than command, but logic would be the same.

ttexalens/parse_elf.py
```diff
@@ -676,18 +676,20 @@ def get_frame_description(self, pc, risc_debug) -> FrameDescription:
 
 
 def decode_symbols(elf_file):
-    functions = {}
+    symbols = {}
```
miacim, line 679: Can you please add changes in this file as part of separate change? These are bugs in parsing. Also add test that  is going to hit this code path.

ttexalens/cli_commands/noc-status.py
```diff
@@ -0,0 +1,98 @@
+# SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
```
miacim, line 1: I had chat with Vuk yesterday. These should be part of the script. Script should have function that can be called externally from other tools like tt_triage, but also you can call it from same file providing optional main. Lets align on this before you continue.

adjordjevic-TT: Added this as a script and put it in ttexalens/scripts folder. Script contains main so it prints inspection results when called from command line. Also, it has check_noc_status function that looks for mismatches and stores them in dictionary and this function can be called externally in order to obtain results from inspection.

ttexalens/scripts/check_noc_status.py
```diff
@@ -0,0 +1,143 @@
+# SPDX-FileCopyrightText: (c) 2025 Tenstorrent AI ULC
```
tt-vjovanovic, line 1: This should probably be checked-in into metal repository as a first script that is using debugging library :)

ttexalens/scripts/check_noc_status.py
```diff
@@ -0,0 +1,143 @@
...
+            error = False
+
+            # Check if variables match with corresponding register
+            for var in VAR_TO_REG_MAP:
+                reg = VAR_TO_REG_MAP[var]
+                address = symbols[var]
+                # If reading fails, write error message and skip to next core
+                try:
+                    reg_val = read_tensix_register(loc, reg, device_id, context)
+                    var_val = read_riscv_memory(loc, address, noc_id, risc_id, device_id, context)
```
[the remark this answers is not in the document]
adjordjevic-TT: Making an issue for this

adjordjevic-TT:
This script is added to tt-metal repo.

PR #354, opened 2025-05-07
Added support for firmware elfs and added object type to decode_symbols

adjordjevic-TT:
Added support for firmware elfs in read_elf by trying do find `.firmware_text` section if `.text` section is not found.

Included object type in return of decode_symbols function.

Also added test_elfs folder in test which contains elf files for testing purposes (additionally it contains .dis and .dump files). For now that folder contains firmware for brisc.


miacim (approved):
Remove dis and dump file from change.

tt-vjovanovic (approved):

PR #367, opened 2025-05-14
Refactor ELF and DWARF handling to use ParsedElfFile for improved clarity and structure

tt-vjovanovic:
Method `read_elf` now returns class instead of dictionary. Fields in class are same as in dictionary, just are lazily loaded. This improves performance since not everything is being used in all scenarios.
New class has been added as well. When we load elf on offset, we can do that on already parsed elf, instead of forcing elf parsing again and again for different offsets.
This will greatly improve performance of tt-triage and make integration more readable.

### DIFF
--- a/ttexalens/parse_elf.py
+++ b/ttexalens/parse_elf.py
@@ -88,10 +88,10 @@ def strip_DW_(s):
     return re.sub(r"^DW_[^_]*_", "", s)
 
 
-class MY_DWARF:
+class ElfDwarf:
     def __init__(self, dwarf: DWARFInfo):
         self.dwarf = dwarf
-        self._cus: Dict[int, MY_CU] = {}
+        self._cus: Dict[int, ElfCompileUnit] = {}
 
     @cached_property
     def range_lists(self):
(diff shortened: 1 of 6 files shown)


ttexalens/parse_elf.py
```diff
@@ -168,14 +173,45 @@ def file_lines_ranges(self):
...
     def find_file_line_by_address(self, address):
-        address += self.loaded_offset
         ranges = self.file_lines_ranges
         for (start, end), info in ranges.items():
             if start <= address < end:
                 return info
         return None
 
 
+class MY_DWARFWithOffset(MY_DWARF):
```
ihamer-tt, line 183: Maybe remove MY_ from the name of this class and MY_CU.

ihamer-tt (approved):

ttexalens/parse_elf.py
tt-vjovanovic: Renamed all classes to remove `MY_` from name

miacim (approved):

PR #372, opened 2025-05-16
Fixed some bugs in callstack

adjordjevic-TT:
Made some changes in callstack in order to fix discovered bugs. Changes are the following:
- Offset argument can be given in hex format too
- Disabled relocations when getting dwarf info
- Path for subprograms is now taken from specification symbol in order to include namespaces
- Paths are now printed instead of names for inlined_functions and subprograms in order to include namespaces, inlined functions print path of their abstract origin since their path contains functions that called them
- Added cached_property linkage_name found in symbol attributes (for now it is not used)
- Added cached_property decl_file_info (for now not used)
- If attributes does not contain address range calculate it from children if possible
- Included namespace in callstack test

### DIFF
--- a/ttexalens/debug_risc.py
+++ b/ttexalens/debug_risc.py
@@ -7,7 +7,7 @@ from dataclasses import dataclass
 from typing import List, Union
 from ttexalens.context import Context
 from ttexalens.coordinate import OnChipCoordinate
-from ttexalens.parse_elf import ParsedElfFile, read_elf
+from ttexalens.parse_elf import read_elf
 from ttexalens.tt_exalens_lib import read_word_from_device, write_words_to_device, read_from_device, write_to_device
 from ttexalens import util as util
 import os
@@ -957,7 +957,7 @@ class RiscLoader:
             not self.risc_debug.is_halted() or self.risc_debug.read_status().is_ebreak_hit
         ), f"RISC at location {self.risc_debug.location} is still halted, but not because of ebreak."
 
-    def _read_elfs(self, elf_paths: List[str], offsets: List[int | None]) -> list[ParsedElfFile]:
+    def _read_elfs(self, elf_paths: List[str], offsets: List[int | None]) -> List[dict]:
         if not isinstance(elf_paths, list):
             elf_paths = [elf_paths]
         offsets = [None] * len(elf_paths) if offsets is None else offsets
@@ -969,9 +969,9 @@ class RiscLoader:
 
         return elfs
 
-    def _find_elf_and_frame_description(self, elfs: list[ParsedElfFile], pc: int):
+    def _find_elf_and_frame_description(self, elfs: List[dict], pc: int):
         for elf in elfs:
-            frame_description = elf.frame_info.get_frame_description(pc, self.risc_debug)
+            frame_description = elf["frame-info"].get_frame_description(pc, self.risc_debug)
             # If we get frame description from elf we return that elf and frame description
             if frame_description is not None:
                 return elf, frame_description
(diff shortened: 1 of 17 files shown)


ttexalens/parse_elf.py
```diff
@@ -754,7 +819,7 @@ def read_elf(file_ifc, elf_file_path, load_address=None):
     if load_address is None:
         load_address = text_sh_address
 
-    dwarf = elf.get_dwarf_info()
+    dwarf = elf.get_dwarf_info(relocate_dwarf_sections=False)
```
tt-vjovanovic, line 822: Please add comment why we have disabled relocations...

ttexalens/parse_elf.py
```diff
@@ -51,6 +51,7 @@
     from elftools.elf.enums import ENUM_ST_INFO_TYPE
     from docopt import docopt
     from tabulate import tabulate
+    import cxxfilt
```
tt-vjovanovic, line 54: Please add `cxxfilt` to `requirements.txt`

ttexalens/parse_elf.py
```diff
@@ -422,6 +429,9 @@ def address(self):
                 # Check if the first opcode is DW_OP_addr (0x03)
                 if block[0] == 0x03:
                     addr = int.from_bytes(block[1:], byteorder="little")
+                elif hasattr(block, "__iter__"):
```
tt-vjovanovic, line 432: Can you check if there is built-in function that parses `DW_AT_location` for us? Similarly to how we found `get_DIE_from_attribute` method...

ttexalens/parse_elf.py
```diff
@@ -480,6 +511,8 @@ def name(self):
         else:
             # We can't figure out the name of this variable. Just give it a name based on the ELF offset.
             name = f"{self.tag}-{hex(self.offset)}"
+
```
tt-vjovanovic, line 514: Remove empty lines

ttexalens/parse_elf.py
```diff
@@ -494,8 +527,40 @@ def address_ranges(self):
         elif "DW_AT_ranges" in self.attributes:
             ranges = self.cu.dwarf.range_lists.get_range_list_at_offset(self.attributes["DW_AT_ranges"].value)
             return [(r.begin_offset, r.end_offset) for r in ranges]
+        else:
+            child_ranges = []
+            for child in self.iter_children():
+                child_range = child.address_ranges
+                if child_range:
+                    if isinstance(child_range, tuple):
```
tt-vjovanovic, line 535: Can it ever be tuple if we are always returning array?
Command `make mypy` should help if you add return type on `address_ranges`... In VSCode you can enable more tests with Pylance by setting `python.analysis.typeCheckingMode` to `standard`...

ttexalens/debug_risc.py
```diff
@@ -977,6 +977,12 @@ def _find_elf_and_frame_description(self, elfs: List[dict], pc: int):
                 return elf, frame_description
 
         return None, None
+    
+    def _get_path_from_attribute(self, die, attribute: str) -> str:
```
tt-vjovanovic, line 981: I would really like to not have this method and have all the logic in `parse_elf.py` :)
Can you move logic that calls this method into `name` property of `function_die`?
Now we have test that will prevent us from function name errors (like not having namespace or having duplicate namespace in debug build)...

tt-vjovanovic (approved):

ttexalens/parse_elf.py
adjordjevic-TT: You are right it can't. Fixed it.

ttexalens/parse_elf.py
adjordjevic-TT: Seems to be covered here https://github.com/tenstorrent/tt-exalens/pull/374.

PR #374, opened 2025-05-18
Creating default mem_reader for mem_access

tt-vjovanovic:
Improving how we get die address using DW_AT_location.
Incorporating changes so get_mem_reader returns correct callable so mem_access can work.

At the moment, agreement with UMD team is to write our own unaligned read/write functions and in the future they will support it with different change. See #375 for more info.

### DIFF
--- a/ttexalens/firmware.py
+++ b/ttexalens/firmware.py
@@ -165,8 +165,8 @@ class ELF:
         address, size and type_die of the variable. If the variable is not found, return None.
         """
 
-        def my_mem_reader(addr, size_bytes):
-            pass
+        def my_mem_reader(addr: int, size_bytes: int, elements_to_read: int) -> list[int]:
+            return []
 
         if mem_reader is None:
             mem_reader = my_mem_reader
--- a/ttexalens/firmware.py
+++ b/ttexalens/firmware.py
@@ -6,7 +6,7 @@ This module is used to represent the firmware
 """
 
 import time
-from typing import Dict, Tuple
+from typing import Callable, Dict, Tuple
 from ttexalens import parse_elf
 from ttexalens import util as util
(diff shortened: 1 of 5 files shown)


ttexalens/server/lib/src/umd_implementation.cpp
```diff
@@ -11,6 +11,43 @@
 
 namespace tt::exalens {
 
+static void read_from_device_reg(tt::umd::Cluster* cluster, void* mem_ptr, chip_id_t chip, tt::umd::CoreCoord core,
```
tt-vjovanovic, line 14: @broskoTT, is this the way to go for unaligned reads?

ttexalens/firmware.py
```diff
@@ -210,13 +210,23 @@ def read_path(self, path_str, mem_reader):
         return data
 
     @staticmethod
-    def get_mem_reader(context, device_id, core_loc):
+    def get_mem_reader(context, device_id, core_loc) -> Callable[[int, int, int], list[int]]:
```
tt-vjovanovic, line 213: We need similar method that is using debug hardware to read private memory.
Registers and other complex locations are future work...

adjordjevic-TT (approved):

ttexalens/server/lib/src/umd_implementation.cpp
tt-vjovanovic: Talked offline. This is the way to go for now.

ttexalens/firmware.py
tt-vjovanovic: Opened https://github.com/tenstorrent/tt-exalens/issues/376...

miacim (approved):

PR #391, opened 2025-05-28
Adding postcode offset to Device class

### LINKED ISSUE #384: Expose NOC address for postcode from ARC
In tt-triage we read postcode from ARC memory through NOC. This address is hardcoded for now. To improve this we should expose it in exalens and use that instead.

adjordjevic-TT:
Closes #384 

Added missing scratch registers and provided description for all of them. Also provided link towards atlassian page with scratch register information.

### DIFF
--- a/ttexalens/device.py
+++ b/ttexalens/device.py
@@ -6,7 +6,7 @@ from abc import abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, replace
 from functools import cached_property
-from typing import Dict, Iterable, Iterator, List, Sequence, Tuple
+from typing import Iterable, Sequence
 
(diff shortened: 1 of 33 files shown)


ttexalens/device.py
```diff
@@ -116,6 +116,7 @@ class Device(TTObject):
     PCI_ARC_ROM_DATA_BASE_ADDR: int = None
     NOC_ARC_ROM_DATA_BASE_ADDR: int = None
     NOC_REGISTER_OFFSET: int = None
+    ARC_POSTCODE_OFFSET: int = None
```
tt-vjovanovic, line 119: Is this supposed to represent arc register?

PR #396, opened 2025-05-31
Extracting device classes into NocBlocks

tt-vjovanovic:
Here is a new way of how we get to device specific information (this is only initial PR that introduces layout, API, etc.)
 `Device` class will expose a way to get `NocBlock` which has all necessary definitions to query/alter/configure data on it.
Similarly to previously introduced `DebugBusSignalStore` this PR introduces `RegisterStore` which allows us to read registers. `RegisterStore` requires two parts to work:
- map of register name to register type + relative address
- method that maps register type to `DeviceAddress`

`DeviceAddress` has couple of fields:
- `private_address` - address that can be used to read using debug hardware
- `noc_address` - address that can be used to read over noc
- `raw_address` - address that can be used to read using `read_raw`
- `noc_id` - Id of the noc that should be used for reading. Some NocBlocks can access noc registers only using selected noc.

`NocBlock` should define hardware that it has (like rics cores, L1, regsiters, debug bus signals, etc.).

Previous design has single class per architecture to rule them all (`WormholeDevice` / `BlackholeDevice` / `QuasarDevice`). This way we can define all components that are available on chip and `Device` class will just expose getting blocks.

I changed `device-summary` command to include call to new API. It served as a first test :)

Please take a look at:
- how API looks like from high level
- If I missed to incorporate changes that were pushed recently to device classes
- names, as I won't like to do renaming after this PR
- file locations, as moving them is similar to renaming


ttexalens/hardware/wormhole/arc_block.py
```diff
@@ -0,0 +1,87 @@
...
+    "ARC_RESET_ARC_UDMIAXI_REGION": ArcResetRegisterDescription(offset=0x10C),
+    "ARC_RESET_SCRATCH0": ArcResetRegisterDescription(offset=0x060),
+    "ARC_RESET_SCRATCH1": ArcResetRegisterDescription(offset=0x064),
+    "ARC_RESET_SCRATCH2": ArcResetRegisterDescription(offset=0x068),
+    "ARC_RESET_SCRATCH3": ArcResetRegisterDescription(offset=0x06C),
+    "ARC_RESET_SCRATCH4": ArcResetRegisterDescription(offset=0x070),
+    "ARC_RESET_SCRATCH5": ArcResetRegisterDescription(offset=0x074),
+    "ARC_CSM_DATA": ArcCsmRegisterDescription(offset=0),
+    "ARC_ROM_DATA": ArcRomRegisterDescription(offset=0),
+}
```
adjordjevic-TT, line 31: Same as for blackhole

ttexalens/hardware/wormhole/dram_block.py
```diff
@@ -0,0 +1,48 @@
...
+class WormholeDramBlock(WormholeNocBlock):
+    def __init__(self, location: OnChipCoordinate):
+        super().__init__(location, block_type="dram")
+
+        self.dram_bank = MemoryBlock(
+            size=2 * 1024 * 1024 * 1024, address=DeviceAddress(private_address=0x00000000, noc_address=0x00000000)
+        )
+
+        if location._noc0_coord[1] % 3 == 1:
+            self.register_store_noc0 = RegisterStore(register_store_location0_noc0_initialization, self.location)
```
adjordjevic-TT, line 41: Maybe add comments explaining why we are using modulo 3

adjordjevic-TT (approved):

miacim (approved):

ttexalens/hardware/wormhole/dram_block.py
tt-vjovanovic: Done

PR #398, opened 2025-06-03
Refactor ARC register access

tt-vjovanovic:
Refactoring code to start using new device API to get arc block and use its register store for accessing arc registers.

### DIFF
--- a/ttexalens/hw/arc/arc.py
+++ b/ttexalens/hw/arc/arc.py
@@ -6,7 +6,7 @@ from ttexalens.util import TTException
 import re
 import os
 
-from ttexalens.tt_exalens_lib_utils import check_context, arc_read, arc_write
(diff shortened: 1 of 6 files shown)


ttexalens/hw/arc/arc.py
```diff
@@ -160,20 +157,14 @@ def load_arc_fw(file_name: str, iccm_id: int, device_id: int, context: Context =
         # Same for womhole and grayskull
```
adjordjevic-TT, line 157: Probably outdated comment

adjordjevic-TT (approved):

ttexalens/hw/arc/arc.py
tt-vjovanovic: Removed

PR #414, opened 2025-06-07
Moving arc telemetry tags to arc block

tt-vjovanovic:
Introducing `ArcBlock` class in hardware. This class will be in charge for all arc functionality.
Removing arc telemetry from device classes and moving it to corresponding arc blocks.


miacim:
@tt-vjovanovic, connect pull request with corresponding issue.

Issue #328, opened 2025-04-16
Add Common search Command Across CLI Commands


miacim:
## Description:

The search pattern currently exists in the tensix-reg.py command. This logic should be extracted and made common across all commands to avoid duplication and ensure consistent behavior.

## Action Items:

Refactor existing search implementation into a shared utility module.
Add search support in debug-bus and noc commands.
Ensure consistent argument parsing and output formatting across commands.
Add tests for the shared search logic.

PR #425, opened 2025-06-09
Add shared search functionality across noc, debug-bus and tensix-reg

iklikovacTT:
#328
Moved the search code from cli_commands/tensix-reg.py to util.py, and also added the function to noc and debug-bus. 
Added a unit test for the code (unit_tests/test_search.py).

### DIFF
--- a/ttexalens/util.py
+++ b/ttexalens/util.py
@@ -901,9 +901,9 @@ def search(strings: list[str], pattern: str, max: str | int = "all") -> list[str
         if max != "all":
             n = int(max)
             if n <= 0:
-                raise ValueError(f'Invalid argument for --max. Expected positive integer, "all" or None, but got {max}')
+                raise ValueError(f'Invalid argument for --max. Expected positive integer or "all", but got {max}')
     except:
-        raise ValueError(f'Invalid argument for --max. Expected positive integer, "all" or None, but got {max}')
+        raise ValueError(f'Invalid argument for --max. Expected positive integer or "all", but got {max}')
     if max is None:
         n = 10
     elif max == "all":


ttexalens/util.py
```diff
@@ -891,3 +892,28 @@ def clr_by_index(idx):
...
+        n = -1
+    else:
+        n = int(n)
+
+    pattern = pattern.lower()
+    results = []
+
+    for s in strings:
+        if n == 0:
+            # WARN(f"{len(data) - n} matches remaining. To see more results, increase the --max value.")
```
adjordjevic-TT, line 912: It seems like a good idea to include warning message if number of matches exceeds value of `n`. You probably should not reference `--max` value here since it is not used in this function. You can print something like increase the number of maximum matches displayed. If you want to reference specific argument this should be handled in command (you can achieve this by returning a flag that signals there are some matches not displayed and then you can add printing message in command).

test/ttexalens/unit_tests/test_search.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_exact(self):
+        result = search(["idk", "man"], "idk")
+        self.assertEqual(result, ["idk"])
+    
+    def test_zero_results(self):
+        result = search(["idk"], "man")
+        self.assertEqual(result, [])
+    
+    def test_partial(self):
+        result = search(self.data, "*re*", "2") # 3 matches
```
miacim, line 38: 2 matches

ttexalens/cli_commands/debug-bus.py
```diff
@@ -117,7 +121,20 @@ def run(cmd_text, context, ui_state: UIState = None):
                 if not debug_bus_signal_store:
                     util.ERROR(f"Device {device._id} at location {loc.to_user_str()} does not have a debug bus.")
                     continue
-                print(debug_bus_signal_store.get_signal_names())
+                names = debug_bus_signal_store.get_signal_names()
+                # Filter if requested
+                names = search(list(names), dopt.args["<pattern>"]) if dopt.args["--search"] else names
```
adjordjevic-TT, line 126: If names are empty we can end command execution here since there are no point in continuing.

test/ttexalens/unit_tests/test_search.py
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+# Unit test for ttexalens/util.py:search.
+import unittest
+from ttexalens.util import search
+from fnmatch import fnmatch
+
+class TestSearch(unittest.TestCase):
+    def setUp(self):
+        self.data = ["test", "testing", "tested", "Tenstorrent", "Tensix",
```
miacim, line 12: can you add numbers and '[', ']' characters and any other possible character to data set. Take a look at debug bus singnals for blackhole.

test/ttexalens/unit_tests/test_search.py
```diff
@@ -0,0 +1,42 @@
...
+
+class TestSearch(unittest.TestCase):
+    def setUp(self):
+        self.data = ["test", "testing", "tested", "Tenstorrent", "Tensix", 
+                     "status", "step", "read", "reset", "terminal emulator", 
+                     "technology", "tech", "tender", "text", "team"]
+
+    def test_default(self):
+        result = search(self.data, "te*")
+        self.assertEqual(len(result), 10) # 11 in total, but search is called with default n
```
miacim, line 18: This test does not verify that found results are starting with te, but only that there are 10 of them. If you want to test only default parameter n, than you should change name of the test. a bit.

ttexalens/cli_commands/debug-bus.py
```diff
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 """
 Usage:
-  debug-bus list-names [-v] [-d <device>] [-l <loc>]
+  debug-bus list-names [-v] [-d <device>] [-l <loc>] [--search <pattern>] [-s]
```
miacim, line 6: Update Description below with explanation of search pattern and -s

ttexalens/util.py
```diff
@@ -891,3 +892,28 @@ def clr_by_index(idx):
 
 def color_text_by_index(text, color_index):
     return f"{clr_by_index(color_index)}{text}{CLR_END}"
+
+
+# Return a list of up to n elements from strings that match the given wildcard pattern. 
+# Defaults to 10 elements, negative values of n or "all" mean all.
+def search(strings: list[str], pattern: str, n = None) -> list[str]:
+    if n is None:
```
miacim, line 900: Default should be all.

ttexalens/util.py
```diff
@@ -891,3 +892,28 @@ def clr_by_index(idx):
...
+    for s in strings:
+        if n == 0:
+            # WARN(f"{len(data) - n} matches remaining. To see more results, increase the --max value.")
+            break
+        
+        if fnmatch(s.lower(), pattern):
+            results.append(s)
+            n -= 1
+
+    return results
```
adjordjevic-TT, line 919: If results are empty we probably should print info/warning message to let user know search was performed but it did not find anything.

ttexalens/cli_commands/tensix-reg.py
```diff
@@ -154,20 +140,20 @@ def run(cmd_text, context, ui_state: UIState = None):
...
                     if max_regs <= 0:
                         raise ValueError(f"Invalid value for max-regs. Expected positive integer, but got {max_regs}")
-                else:
-                    max_regs = len(register_names)
             except:
                 raise ValueError(f"Invalid value for max-regs. Expected an integer, but got {max_regs}")
 
-            INFO(f"Register names that match pattern on device {device.id()}")
-            print_matches(register_pattern, register_names, max_regs)
+            results = search(device._get_tensix_register_map_keys(), register_pattern, max_regs)
```
adjordjevic-TT, line 153: If results are empty we can continue to next core here. In my opinion, it is better to print warning message that we did not find anything than it is to print something like Matches: followed by nothing (this can seem like a bug).

test/ttexalens/unit_tests/test_search.py
```diff
@@ -0,0 +1,42 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+# Unit test for ttexalens/util.py:search.
+import unittest
+from ttexalens.util import search
+from fnmatch import fnmatch
+
+class TestSearch(unittest.TestCase):
```
adjordjevic-TT, line 10: You should add some test cases that trigger exceptions. For example int(n) should raise an exception if a string that is not a number is passed as `n`. You can check for exceptions with `assertRaises`.

test/ttexalens/unit_tests/test_search.py
```diff
@@ -0,0 +1,42 @@
...
+                     "status", "step", "read", "reset", "terminal emulator", 
+                     "technology", "tech", "tender", "text", "team"]
+
+    def test_default(self):
+        result = search(self.data, "te*")
+        self.assertEqual(len(result), 10) # 11 in total, but search is called with default n
+
+    def test_max(self):
+        result = search(self.data, "te*", n = "max")
+        expected = [s for s in self.data if fnmatch(s.lower(), "te*")]
```
tt-vjovanovic, line 22: Usually practice is not to trust method of testing :)
You wrote `search` method using `fnmatch` and now you are testing match with `fnmatch`...
What would be the best is to hand craft expected results based on your input array. What would user expect to get as a result, not some other function :)
Similar to what you used in `test_exact` and `test_zero_results`....
I assume it is because of `te*` testing, you wanted to include many results to test `max` argument, but for testing `max` argument, you should use simple pattern like `*` and not worry about search itself...

ttexalens/util.py
iklikovacTT: Done. I left `None` as a legal value so that it can be passed directly from `dopt.args` by the caller.

ttexalens/cli_commands/debug-bus.py
```diff
@@ -3,12 +3,18 @@
...
 """
 Usage:
-  debug-bus list-names [-v] [-d <device>] [-l <loc>]
+  debug-bus list-names [-v] [-d <device>] [-l <loc>] [--search <pattern>] [--max <max-sigs>] [-s]
   debug-bus [<signals>] [-v] [-d <device>] [-l <loc>]
 
+Options:
+    -s, --simple            Print simple output
+    --search <pattern>      Search for signals by pattern (in wildcard format)
+    --max <max-sigs>        Limit --search output (default: 10, use --max "all" to print all matches)
```
miacim, line 12: is limit only on search output or generally on list-names output

test/ttexalens/unit_tests/test_search.py
```diff
@@ -0,0 +1,96 @@
...
+            "O_RDWR",
+            "__builtin_unreachable",
+            "BLACKHOLE",
+            "<thing>",
+            "42",
+            "_3",
+        ]
+
+    def test_default_n(self):
+        result = search(self.data, "*")
```
miacim, line 39: looks like you are missing assert

ttexalens/cli_commands/debug-bus.py
iklikovacTT: Only on --search currently, though I could change that

test/ttexalens/unit_tests/test_search.py
iklikovacTT: Right, removed that test since it's wholly covered by `test_max` now that `all` is default.

ttexalens/util.py
```diff
@@ -891,3 +892,32 @@ def clr_by_index(idx):
 
 def color_text_by_index(text, color_index):
     return f"{clr_by_index(color_index)}{text}{CLR_END}"
+
+
+# Return a list of up to n elements from strings that match the given wildcard pattern.
+# Defaults to 10 elements, negative values of max or "all" mean all.
+def search(strings: list[str], pattern: str, max: str | int | None = "all") -> list[str]:
```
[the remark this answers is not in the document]
adjordjevic-TT: This is usually handled in command, something like this:
```
max = dopt.args["<max>"] if dopt.args["<max>"] else "all"
```
And then you pass max as argument for search.

Issue #280, opened 2025-03-11
Add test if noc1 implementation is working correctly


tt-vjovanovic:
Start tt-exalens with noc1 context. Get noc1 counter on some core. Do some transactions (read/write on that core). Check that noc1 counter is increased by at least amount of transactions.

tt-vjovanovic:
Blocked by https://github.com/tenstorrent/tt-exalens/issues/263

miacim:
We have exposed noc registers that can be used to add test. noc status command is printing these register. Issue is not blocked on #263.

miacim:
#455 test reveals issue on blackhole.

PR #445, opened 2025-06-13
Add NOC tests

iklikovacTT:
#280 
Added `init_test_context` into `test_base.py` that starts a TTExaLens context on either NOC0 or NOC1.
Added `test_noc.py` that checks if the NOCs behave consistently regardless of which one hosts the context. 
The test sets up two contexts (NOC0 and NOC1), and for each context:
- reads the transaction counters,
- performs differently-sized writes through NOC0 and NOC1,
- and checks that each NOC counter increased by at least the amount corresponding to the data sent through it.


test/ttexalens/unit_tests/test_base.py
```diff
@@ -12,3 +12,11 @@ def init_default_test_context():
         return tt_exalens_init.init_ttexalens_remote(ip_address, port)
     else:
         return tt_exalens_init.init_ttexalens()
+
+
+def init_test_context(use_noc1: bool = True):
```
miacim, line 17: To be consistent, default should be noc0

tt-vjovanovic (approved):

PR #450, opened 2025-06-14
Moving RiscLoader to new API

tt-vjovanovic:
Moving `RiscLoader` to new file (`risc_loader.py`).
Replacing usage of `RiscLoader` with new class.
Many small fixes along the way...


test/ttexalens/unit_tests/test_risc_debug.py
```diff
@@ -112,7 +113,7 @@ def test_read_write_gpr(self):
         if not (self.core_sim.is_blackhole() and self.core_sim.is_eth_block()):
             # PC is not readable through GPR on blackhole ETH core for now
             self.assertEqual(
-                self.core_sim.read_gpr(get_register_index("pc")),
+                self.core_sim.get_pc(),
```
adjordjevic-TT, line 116: Since here we are testing read/write GPR, does it make sense to use `get_pc` since it reads PC through debug bus?

test/ttexalens/unit_tests/test_risc_debug.py
```diff
@@ -112,7 +113,7 @@ def test_read_write_gpr(self):
...
                 self.core_sim.program_base_address + 4,
                 "PC should be 4.",
```
adjordjevic-TT, line 118: PC should be `self.core_sim.program_base_address + 4`

adjordjevic-TT (approved):

PR #458, opened 2025-06-18
Removing debug_risc.py in favor of new API

tt-vjovanovic:
Removing remaining places of `debug_risc.py`.
Changing `--risc` to accept and return `risc_name` instead of `risc_id`.
`device-summary` now prints status of all risc cores on noc_block (if there are any).
Updating `parse_elf` to use new `RiscDebug`.


ttexalens/cli_commands/callstack.py
```diff
@@ -13,11 +13,11 @@
 
 Options:
   -o <offsets>        List of offsets for each elf file, comma separated.
-  -r <risc>           RiscV ID (0: brisc, 1-3 triscs). [Default: 0]
+  -r <risc>           RiscV name (brisc, triscs0, triscs1, triscs2, ncrisc, erisc). [default: first risc]
```
adjordjevic-TT, line 16: 'erisc0' and 'erisc1' are also viable names if you wanted to list them all. We should probably add option to list possible risc names depending on core location.

adjordjevic-TT, line 16: Should `ncrisc` be listed as possible argument?

adjordjevic-TT (approved):

ttexalens/cli_commands/callstack.py
tt-vjovanovic: This was list of examples... Might have forgotten `e.g.` :)

Issue #364, opened 2025-05-12
Remove write_soc_descriptor and start using tt_cluster_descriptor::serialize()


tt-vjovanovic:
There is new function in UMD `tt_cluster_descriptor::serialize()`. It is not complete, but we should start using it and contribute with what we need.

`write_soc_descriptor` was copied from buda repository. It should be removed and all missing printouts that we have today should be added to `tt_cluster_descriptor::serialize()`.

adjordjevic-TT:
Pending investigation. Latest UMD breaks device after running our tests. This happened with commit before commit that introduced serialization of soc descriptor.

tt-vjovanovic:
Related to https://github.com/tenstorrent/tt-exalens/issues/26

PR #460, opened 2025-06-18
Adding types of ETH cores (active/idle) in device description yaml

adjordjevic-TT:
Now when generating `device_desc_runtime.yaml` we include information about types of ETH cores (active or idle). For this we use existing functionality from UMD.

In device description we still keep `eth` field that contains locations of all ETH cores, but add 2 additional fields (`idle_eth` and `active_eth`) that contain locations of idle and active ETH cores.

We are now able, with `get_block(block_type)` to obtain only active ETH cores (`block_type="active_eth"`) or only idle (`block_type="idle_eth"`). We still can get all ETH cores with `block_type="eth"`.

### DIFF
--- a/ttexalens/device.py
+++ b/ttexalens/device.py
@@ -299,6 +299,8 @@ class Device(TTObject):
             "color": util.CLR_GREEN,
         },
         "eth": {"symbol": "E", "desc": "Ethernet", "core_type": "eth", "color": util.CLR_YELLOW},
+        "idle_eth": {"symbol": "E", "desc": "Idle Ethernet", "core_type": "eth", "color": util.CLR_YELLOW},
+        "active_eth": {"symbol": "E", "desc": "Active Ethernet", "core_type": "eth", "color": util.CLR_YELLOW},
         "arc": {"symbol": "A", "desc": "ARC", "core_type": "arc", "color": util.CLR_GREY},
         "dram": {"symbol": "D", "desc": "DRAM", "core_type": "dram", "color": util.CLR_TEAL},
         "pcie": {"symbol": "P", "desc": "PCIE", "core_type": "pcie", "color": util.CLR_GREY},
--- a/ttexalens/hw/tensix/blackhole/blackhole.py
+++ b/ttexalens/hw/tensix/blackhole/blackhole.py
@@ -680,6 +680,10 @@ class BlackholeDevice(Device):
             return BlackholeDramBlock(location)
         elif block_type == "eth":
             return BlackholeEthBlock(location)
+        elif block_type == "idle_eth":
+            return BlackholeEthBlock(location)
+        elif block_type == "active_eth":
+            return BlackholeEthBlock(location)
         elif block_type == "functional_workers":
             return BlackholeFunctionalWorkerBlock(location)
         elif block_type == "harvested_workers":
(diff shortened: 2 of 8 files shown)


ttexalens/hw/tensix/blackhole/blackhole.py
```diff
@@ -680,6 +680,10 @@ def get_block(self, location):
             return BlackholeDramBlock(location)
         elif block_type == "eth":
             return BlackholeEthBlock(location)
+        elif block_type == "idle_eth":
```
tt-vjovanovic, line 683: Idea is that block_type is unique for location. That will not be the case with new yaml file. Adding info to soc_descriptor yaml file doesn't require it to be part of block_type :) For example `worker_l1_size`...
We should look for a way how to expose data from yaml file to the end user. Probably just two simple functions `get_idle_eth_locations` and `get_active_eth_locations` in device class...

ttexalens/server/lib/src/open_implementation.cpp
```diff
@@ -123,9 +123,11 @@ static void write_coord(std::ostream &out, const tt_xy_pair &xy, CoreType core_t
 // Creates SOC descriptor files by serializing tt_SocDescroptor structure to yaml.
 // TODO: Current copied from runtime/runtime_utils.cpp: print_device_description. It should be moved to UMD and reused
 // on both places.
-static void write_soc_descriptor(std::string file_name, const tt_SocDescriptor &soc_descriptor) {
+static void write_soc_descriptor(std::string file_name, const tt_SocDescriptor &soc_descriptor, uint8_t device_id) {
```
tt-vjovanovic, line 126: I know it is not part of this change, but while you are changing this function, consider doing this change as well :)
https://github.com/tenstorrent/tt-exalens/issues/364

ttexalens/device.py
```diff
@@ -262,6 +262,25 @@ def arc_block(self) -> ArcBlock:
...
 
+    @cached_property
+    def active_eth_blocks(self) -> list[NocBlock]:
+        active_channels = []
+        for connection in self.cluster_desc["ethernet_connections"]:
+            for endpoint in connection:
+                if endpoint["chip"] == self._id:
+                    active_channels.append(endpoint["chan"])
+
+        return [self.get_blocks(block_type="eth")[chan] for chan in active_channels]
```
tt-vjovanovic, line 273: Not sure how good optimization of caches it is. Can you change code to execute get_blocks method only once and to create single list object instead of creating list of ints and converting that to list on NocBlocks...

ttexalens/device.py
```diff
@@ -262,6 +262,25 @@ def arc_block(self) -> ArcBlock:
...
+
+    @cached_property
+    def idle_eth_blocks(self) -> list[NocBlock]:
+        idle_blocks = []
+        for block in self.get_blocks(block_type="eth"):
+            if not block in self.active_eth_blocks:
+                idle_blocks.append(block)
```
tt-vjovanovic, line 280: I'm afraid that caches might not be setup correctly either now or if someone changes that in the future... Currently it should do verification that it is the same object instance, but in the future could be changed to operator equals (or whatever is Python concept)...
Can you add test that does simple  verification that all locations of returned blocks for idle and active eths are inside eth blocks and that there is no overlap between them...

tt-vjovanovic (approved):

ttexalens/device.py
adjordjevic-TT: Added `test_device.py`, for now we are only testing this. Do you have some ideas of what else should be tested regarding `Device` class?

PR #490, opened 2025-07-14
Allow printing actual core types from gdb

adjordjevic-TT:
In `gdb_server.py` added logic to get actual core type (previously we assigned `worker` to every process, but now we have both eth and tensix cores so this is no longer viable).

To enable this added `get_block_type` function to `RiscDebug` that returns `block_type` from `NocBlock` of that risc.

### DIFF
--- a/ttexalens/hardware/risc_debug.py
+++ b/ttexalens/hardware/risc_debug.py
@@ -74,11 +74,6 @@ class RiscDebug:
         noc_block = risc_location.location._device.get_block(risc_location.location)
         return noc_block.get_risc_debug(risc_location.risc_name, risc_location.neo_id)
 
-    @staticmethod
-    def get_block_type(risc_location: RiscLocation) -> str:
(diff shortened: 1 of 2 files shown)


ttexalens/hardware/risc_debug.py
```diff
@@ -74,6 +74,11 @@ def get_instance(risc_location: RiscLocation) -> "RiscDebug":
         noc_block = risc_location.location._device.get_block(risc_location.location)
         return noc_block.get_risc_debug(risc_location.risc_name, risc_location.neo_id)
 
+    @staticmethod
```
tt-vjovanovic, line 77: Please remove this method and use Device.get_block_type

tt-vjovanovic: Consider adding maybe "shortcut" method to OnChipCoordinate that would not be static and would return block type...

tt-vjovanovic (approved):

PR #492, opened 2025-07-14
Fixing bug that causes attach to target multiple processes

### LINKED ISSUE #491: GDB attach targeting multiple processes
When trying to attach to a process using `GDB` it attaches to process with given `pid` but also to a process with pid that seems to be hex value of given process. For example, if given pid is 11, gdb attaches to that process, but also process with pid=17.

There is most likely a bug in `gdb_server.py` or 'gdb_communication.py`. 

```
(gdb) info inferior
  Num  Description       Connection                         Executable        
* 1    <null>            1 (extended-remote localhost:6767)                   
(gdb) attach 11
Attaching to process 11
warning: [unknown_elf_file_path]: No such file or directory.
[New inferior 2]
warning: [unknown_elf_file_path]: No such file or directory.
[New Thread 17.17]
[Switching to Thread 17.17]
Truncated register 16 in remote 'g' packet
(gdb) info inferior
  Num  Description       Connection                         Executable        
* 1    process 11        1 (extended-remote localhost:6767)                   
  2    process 17        1 (extended-remote localhost:6767)                   
```

adjordjevic-TT:
Closes #491 

Fixes bug that caused attach to target multiple processes.

We were sending process and thread ids as hex in one place (which is correct) and as decimal in other (which caused a bug).

### DIFF
--- a/ttexalens/gdb/gdb_data.py
+++ b/ttexalens/gdb/gdb_data.py
@@ -12,7 +12,7 @@ class GdbThreadId:
     thread_id: int
 
     def to_gdb_string(self):
-        return f"p{self.process_id}.{self.thread_id}"
+        return f"p{hex(self.process_id)[2:]}.{hex(self.thread_id)[2:]}"
 
 
 @dataclass


ttexalens/gdb/gdb_data.py
```diff
@@ -12,7 +12,7 @@ class GdbThreadId:
     thread_id: int
 
     def to_gdb_string(self):
-        return f"p{self.process_id}.{self.thread_id}"
+        return f"p{hex(self.process_id)[2:]}.{hex(self.thread_id)[2:]}"
```
tt-vjovanovic, line 15: Just check if there is a format type like :x instead of using hex and a substring of it...

PR #493, opened 2025-07-16
Better handling of unknown process error

adjordjevic-TT:
After issuing attach command before sending correct packet (`Hg<pid>.<thread_id>`) gdb sends `Hg0.0` so every time we use attach we print error. This PR stops error printing in `H` command if both process and thread ids are 0.

### DIFF
--- a/ttexalens/gdb/gdb_server.py
+++ b/ttexalens/gdb/gdb_server.py
@@ -297,6 +297,11 @@ class GdbServer(threading.Thread):
                     # We are not changing current process
                     thread_id = self.current_process.thread_id
                 else:
+                    # This is expected behavior
+                    if thread_id.process_id == 0 and thread_id.thread_id == 0:
+                        writer.append(b"E01")
+                        return True
+
                     # Since our processes have only 1 thread, we will pick this up from the cache
                     t = self.debugging_threads.get(thread_id.process_id)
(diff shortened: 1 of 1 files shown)


ttexalens/gdb/gdb_server.py
```diff
@@ -300,8 +300,11 @@ def process_message(self, parser: GdbMessageParser, writer: GdbMessageWriter):
                     # Since our processes have only 1 thread, we will pick this up from the cache
                     t = self.debugging_threads.get(thread_id.process_id)
                     if t is None:
-                        # Unknown process!!!
-                        util.ERROR(f"GDB: Unknown process id in self.debugging_threads: {thread_id.process_id}")
+                        # Process and thread ids equal to 0 is expected behaivour so we should not print error
+                        # Every time before sending message with actual pid/thread id gdb sends Hgp0.0
+                        if not (thread_id.process_id == 0 and thread_id.thread_id == 0):
+                            # Unknown process!!!
+                            util.ERROR(f"GDB: Unknown process id in self.debugging_threads: {thread_id.process_id}")
```
miacim, line 307: If this is not expected behavior, shouldn't we disconnect session

miacim (approved):

miacim (approved):

PR #498, opened 2025-07-22
CoordSystem from PHYSICAL to NOC0

broskoTT:
Part of effort: https://github.com/tenstorrent/tt-umd/issues/621


ttexalens/server/lib/src/open_implementation.cpp
```diff
@@ -544,7 +544,7 @@ std::optional<std::tuple<uint8_t, uint8_t>> open_implementation<BaseClass>::conv
     if (coord_system == "logical") {
         coord_system_enum = CoordSystem::LOGICAL;
     } else if (coord_system == "physical") {
-        coord_system_enum = CoordSystem::PHYSICAL;
```
tt-vjovanovic, line 547: If physical coordinate system is removed from UMD, we should also remove it from our code. It was added only for people that used that in UMD and wanted to have similar experience...

PR #533, opened 2025-08-01
Enabling reading 32 bit formats from dest on blackhole

### LINKED ISSUE #517: Use direct access on blackhole to read dest
Blackhole enables direct access to dest register (0xFFBD8000). It simplifies algorithm for reading dest, and currently probably the only correct solution for reading dest in Int32 and Float32.

adjordjevic-TT:
Closes #517 

Implemented reading dest directly from memory (`0xFFBD8000`) for 32 bit formats (`int32` and `float32`). In this case to get correct values reordering is not necessary (`int32` is printed as-is, `float32` is just unpacked).

Added option for number of tiles to print in `dump_regfile` command. We also now print `TILE ID` between tiles.

### DIFF
--- a/ttexalens/debug_tensix.py
+++ b/ttexalens/debug_tensix.py
@@ -9,7 +9,7 @@ from ttexalens.context import Context
 from ttexalens.hw.tensix.wormhole.wormhole import WormholeDevice
 from ttexalens.hw.tensix.blackhole.blackhole import BlackholeDevice
 from ttexalens.register_store import RegisterDescription
-from ttexalens.tt_exalens_lib import check_context, validate_device_id, read_word_from_device
+from ttexalens.tt_exalens_lib import check_context, validate_device_id
 from ttexalens.util import WARN, TTException
 from ttexalens.device import Device
 from ttexalens.unpack_regfile import unpack_data, TensixDataFormat
@@ -29,6 +29,9 @@ class REGFILE(Enum):
     SRCA = 0
     SRCB = 1
     DSTACC = 2
+
+
+TILE_SIZE = 32 * 32
 
 
 def convert_regfile(regfile: int | str | REGFILE) -> REGFILE:
@@ -54,6 +57,7 @@ class TensixDebug:
     device_id: int
     context: Context
     device: Device
+    dest_start_address: int
 
     def __init__(
         self,
@@ -71,6 +75,7 @@ class TensixDebug:
             self.core_loc = OnChipCoordinate.create(core_loc, device=self.device)
         else:
             self.core_loc = core_loc
+        self.dest_start_address = 0xFFBD8000 if type(self.device) == BlackholeDevice else None
 
     def dbg_buff_status(self):
         return self.register_store.read_register("RISCV_DEBUG_REG_DBG_INSTRN_BUF_STATUS")
@@ -160,6 +165,26 @@ class TensixDebug:
         """
         self.register_store.write_register(register, value)
 
+    def direct_dest_read(self, df: TensixDataFormat, num_tiles: int = 1) -> list[int | float]:
+        if self.dest_start_address is None:
+            raise TTException("Direct dest reading not supported for this architecture.")
+
+        data = []
+        # Using TRISC0 debug hardware to read memory
+        risc_debug = self.noc_block.get_risc_debug(risc_name="trisc0")
+        # TODO: Add option to print multiple tiles
+        for i in range(TILE_SIZE):
+            address = self.dest_start_address + 4 * i
+            with risc_debug.ensure_halted():
+                rd_data = risc_debug.read_memory(address)
+            # Interpreting data as float32
+            if df == TensixDataFormat.Float32:
+                rd_data = struct.unpack(">f", rd_data.to_bytes(4, "big"))[0]
+
+            data.append(rd_data)
+
+        return data
+
     def read_regfile_data(
         self,
         regfile: int | str | REGFILE,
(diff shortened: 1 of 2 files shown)


ttexalens/debug_tensix.py
```diff
@@ -29,6 +31,9 @@ class REGFILE(Enum):
     DSTACC = 2
 
 
+TILE_SIZE = 32 * 32
```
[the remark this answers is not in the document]
miacim: It is fixed.

ttexalens/debug_tensix.py
```diff
@@ -222,24 +257,32 @@ def read_regfile_data(
...
 
-    def read_regfile(self, regfile: int | str | REGFILE) -> list[int | float | str]:
+    def read_regfile(self, regfile: int | str | REGFILE, num_tiles: int | None = None) -> list[int | float | str]:
         """Dumps SRCA/DSTACC register file from the specified core, and parses the data into a list of values.
         Dumping DSTACC on Wormhole as FP32 clobbers the register.
 
         Args:
                 regfile (int | str | REGFILE): Register file to dump (0: SRCA, 1: SRCB, 2: DSTACC).
-
+                num_tiles (int | None): Number of tiles to print. Only available for direct dest read on blackhole. Default: None
```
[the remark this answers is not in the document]
miacim: Yes, we should be able to read more tiles on wormhole also.

ttexalens/debug_tensix.py
```diff
@@ -52,6 +57,7 @@ class TensixDebug:
     device_id: int
     context: Context
     device: Device
+    dest_start_address: int
```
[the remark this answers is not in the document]
adjordjevic-TT: This address is not part of private memory block, but we read it through debug hardware. Should address be added as `private_address`, `noc_address` or `raw_address`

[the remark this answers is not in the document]
tt-vjovanovic: `private_address` is something that can be accessed using private debug hardware (address which can be used from risc cores from within the noc block).
`noc_address` is address available over noc.
`raw_address` is address that can be read using `raw_*` API, usually only PCIE and ARC noc blocks...

ttexalens/debug_tensix.py
```diff
@@ -222,24 +257,32 @@ def read_regfile_data(
...
         """
         regfile = convert_regfile(regfile)
-        df = self.read_tensix_register("ALU_FORMAT_SPEC_REG2_Dstacc")
+        df = TensixDataFormat(self.read_tensix_register("ALU_FORMAT_SPEC_REG2_Dstacc"))
+
+        if regfile == REGFILE.DSTACC and not self._direct_dest_read_enabled(df) and num_tiles is not None:
+            WARN("num_tiles argument only has effect for 32 bit formats on blackhole.")
 
+        # Directly reading dest does not require complex unpacking so we return it right away
+        if regfile == REGFILE.DSTACC and self._direct_dest_read_enabled(df):
```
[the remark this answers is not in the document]
adjordjevic-TT: The idea was that `read_regfile_data` returns data that was not unpacked, and `direct_dest_read` returns unpacked that, but I can make this logic a part of `read_regfile_data` if it makes more sense to you.

ttexalens/debug_tensix.py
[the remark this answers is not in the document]
adjordjevic-TT: Made an issue https://github.com/tenstorrent/tt-exalens/issues/540 for this

ttexalens/debug_tensix.py
```diff
@@ -256,17 +308,24 @@ def read_regfile(self, regfile: int | str | REGFILE) -> list[int | float | str]:
...
+                # If we use dircet read data is already unpacked
+                return data
+            else:
+                return unpack_data(data, df)
         except ValueError as e:
             # If the data format is unsupported, return the raw data.
             WARN(e)
-            WARN("Printing raw data...")
-            return [hex(datum) for datum in data]
+            raw_data: list[str] = []
```
fvranicTT, line 326: Nit: you could use list comprehension here:
```python
raw_data: list[str] = [hex(datum) for datum in data if isinstance(datum, int)]
```
It’s more idiomatic and could be faster.

ttexalens/unpack_regfile.py
```diff
@@ -139,6 +139,17 @@ def unpack_fp32(data) -> list[float]:
     return floats
 
 
+def reorder_uint16(data: list[int]) -> list[int]:
+    for i in range(0, len(data), 2):
+        data[i], data[i + 1] = data[i + 1], data[i]
+    return data
+
+
+def unpack_uint16(data: list[int]) -> list[int]:
```
fvranicTT, line 148: Would this work?
https://github.com/tenstorrent/tt-llk/blob/e1322dd9e5fd0ad416a82adc5ac0527abfe9b641/tests/python_tests/helpers/unpack.py#L45

In your original code, it looks like you are reading bytes in big-endian format and then compensating for it by using a helper method to swap the byte order. On the other hand, it’s possible that our approach is incorrect because we are unpacking under the wrong assumption about the data’s endianness.
FYI @skotaracTT

adjordjevic-TT: This does not seem to work. When I read bytes each integer is in big endian format but it is swapped with integer next to it so it requires to be read as big endian but also needs swapping afterwards (this is assuming your writing is correct since I have been using your tests to test my code).
When I read 1 word I get something like this `[MSB[i+1], LSB[i+1], MSB[i], LSB[i]], after unpacking it gets converted to [int[i+1], int[i]] and then after swapping we get expected [int[i], int[i+1]].

ttexalens/debug_tensix.py
```diff
@@ -159,15 +162,72 @@ def write_tensix_register(self, register: str | RegisterDescription, value: int)
...
+        # 8 bit integer formats are written in 32 bit mode so we can read them directly
+        return type(self.device) == BlackholeDevice and (self._is_32_bit_format(df) or self._is_8_bit_int_format(df))
+
+    @staticmethod
+    def _unpack_value(value: int, df: TensixDataFormat) -> int | float:
+        if df == TensixDataFormat.Float32:
+            return struct.unpack(">f", value.to_bytes(4, "big"))[0]
+        # Because in ALU register format for dest is INT32 for both INT32 and UINT32 we need to check sign bit
+        elif df == TensixDataFormat.Int32:
+            if (value & 0x80000000) == 0x80000000:
```
fvranicTT, line 196: No need for == … part. It will be true if sign bit is set.

ttexalens/debug_tensix.py
```diff
@@ -159,15 +162,72 @@ def write_tensix_register(self, register: str | RegisterDescription, value: int)
...
+                return struct.unpack(">i", value.to_bytes(4, "big"))[0]
+            else:
+                # UINT32 case
+                return value
+        # Same as for INT32/UINT32
+        elif df == TensixDataFormat.Int8:
+            if (value & 0x80000000) == 0x80000000:
```
fvranicTT, line 203: Ditto

ttexalens/debug_tensix.py
```diff
@@ -159,15 +162,54 @@ def write_tensix_register(self, register: str | RegisterDescription, value: int)
...
+            WARN(
+                f"Number of tiles given {num_tiles} is not valid, defaulting to maximum number of tiles {max_num_tiles}"
+            )
+            return max_num_tiles
+
+        return num_tiles
+
+    @staticmethod
+    def _is_32_bit_format(df: TensixDataFormat) -> bool:
+        return df in (TensixDataFormat.Float32, TensixDataFormat.Int32)
```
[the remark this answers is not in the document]
fvranicTT: Should it then be checked here explicitly as well?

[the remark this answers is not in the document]
adjordjevic-TT: Yeah, thanks!
Realistically, `df` will never be `UInt32` because when you deal with `UInt32` you write `Int32` to ALU register that determines dest format. Same is happening for `Int8/UInt8` but I will add it here so we do not regress in the future if something changes.

ttexalens/unpack_regfile.py
```diff
@@ -139,6 +140,17 @@ def unpack_fp32(data) -> list[float]:
...
```
fvranicTT, line 149: Can we do this in one sweep, without a helper method?
```python
def unpack_uint16(data: list[int]) -> list[int]:
    byte_data = bytes(data)
    unpacked = struct.unpack(f">{len(byte_data)//2}H", byte_data)

    # Reorder in pairs
    result = []
    for i in range(0, len(unpacked), 2):
        if i + 1 < len(unpacked):
            result.extend([unpacked[i + 1], unpacked[i]])
        else:
            result.append(unpacked[i])

    return result
```
In this version, we're doing reordering on the fly :)

ttexalens/debug_tensix.py
```diff
@@ -188,60 +242,69 @@ def read_regfile_data(
...
+        df = TensixDataFormat(self.read_tensix_register("ALU_FORMAT_SPEC_REG2_Dstacc"))
+
+        if regfile == REGFILE.DSTACC and not self._direct_dest_read_enabled(df) and num_tiles is not None:
+            WARN("num_tiles argument only has effect for 32 bit formats on blackhole.")
 
+        data: list[int | float] = []
         # Workaround for an architectural quirk of Wormhole: reading DST as INT32 or FP32
         # returns zeros on the lower 16 bits of each datum. This handles the FP32 case.
-        if regfile == REGFILE.DSTACC and df == 0 and type(self.device) == WormholeDevice:
+        if regfile == REGFILE.DSTACC and df == TensixDataFormat.Float32 and type(self.device) == WormholeDevice:
```
fvranicTT, line 305:
```suggestion
        if regfile == REGFILE.DSTACC and df == TensixDataFormat.Float32 and isinstance(self.device, WormholeDevice):
```
ditto

fvranicTT:
How hard is to write tests for these new features @adjordjevic-TT ?
Code LGTM, but it'd be great if we could add a few tests.

adjordjevic-TT:
> How hard is to write tests for these new features @adjordjevic-TT ? Code LGTM, but it'd be great if we could add a few tests.

The problem is we currently do not have the way of writing to dest so before we add that testing is not possible. I have already made an [issue](https://github.com/tenstorrent/tt-exalens/issues/543) but I doubt I have time to resolve it before my vacation so it will probably be best to handle that in separate PR.

fvranicTT:
> > How hard is to write tests for these new features @adjordjevic-TT ? Code LGTM, but it'd be great if we could add a few tests.
> 
> The problem is we currently do not have the way of writing to dest so before we add that testing is not possible. I have already made an [issue](https://github.com/tenstorrent/tt-exalens/issues/543) but I doubt I have time to resolve it before my vacation so it will probably be best to handle that in separate PR.

How are we testing these methods then? 😁

ttexalens/debug_tensix.py
```diff
@@ -188,60 +242,69 @@ def read_regfile_data(
...
+            num_tiles = self._validate_number_of_tiles(num_tiles)
+            data = self.direct_dest_read(df, num_tiles)
+        else:
+            self.register_store.write_register("RISCV_DEBUG_REG_DBG_ARRAY_RD_EN", 1)
+            for row in range(64):
+                row_addr = row if regfile != REGFILE.SRCA else 0
+                regfile_id = 2 if regfile == REGFILE.SRCA else regfile.value
+
+                if regfile == REGFILE.SRCA:
+                    self.inject_instruction(ops.TT_OP_SFPLOAD(3, 0, 0, 0), thread_id)
```
fvranicTT, line 257: We could avoid a lot of repetition here, like:
```python
srca_ops = [
    ops.TT_OP_SFPLOAD(3, 0, 0, 0),
    ops.TT_OP_SFPLOAD(3, 0, 0, 2),
    ops.TT_OP_STALLWAIT(0x40, 0x4000)
]
for op in common_ops:
    self.inject_instruction(op, thread_id)
```

And similar for srcB. That way, code seems less verbose.

ttexalens/debug_tensix.py
```diff
@@ -188,60 +242,69 @@ def read_regfile_data(
...
+                elif regfile == REGFILE.SRCB:
+                    self.inject_instruction(ops.TT_OP_SETRWC(0, 0, 0, 0, 0, 0xF), thread_id)
+                    self.inject_instruction(ops.TT_OP_SETDVALID(0b10), thread_id)
+                    self.inject_instruction(ops.TT_OP_CLEARDVALID(0b10, 0), thread_id)
+                    self.inject_instruction(ops.TT_OP_SETDVALID(0b10), thread_id)
+                    self.inject_instruction(ops.TT_OP_SHIFTXB(7, 0, row_addr), thread_id)
+                    self.inject_instruction(ops.TT_OP_CLEARDVALID(0b10, 0), thread_id)
+
+                for i in range(8):
+                    dbg_array_rd_cmd = (row_addr) + (i << 12) + (regfile_id << 16)
```
fvranicTT, line 270: No need to repetitively calculate regfile_id since it doesn't change inside the loop. We could precompute it and just add the moving part inside the loop.
```python
base_cmd = row_addr + (regfile_id << 16)
for i in range(8):
    dbg_array_rd_cmd = base_cmd + (i << 12)
    ...
```

ttexalens/debug_tensix.py
```diff
@@ -159,15 +162,66 @@ def write_tensix_register(self, register: str | RegisterDescription, value: int)
...
+            raise TTException(f"Unsupported data format {df} for unpacking.")
+
+    def direct_dest_read(self, df: TensixDataFormat, num_tiles: int) -> list[int | float]:
+        if not self._direct_dest_read_enabled(df):
+            raise TTException("Direct dest reading not supported for this architecture or data format.")
+
+        data: list[int | float] = []
+        # Using TRISC0 debug hardware to read memory
+        risc_debug = self.noc_block.get_risc_debug(risc_name="trisc0")
+        for i in range(num_tiles * TILE_SIZE):
```
fvranicTT, line 212: Would this work as well?
```python
if isinstance(self.noc_block, BlackholeFunctionalWorkerBlock):
        risc_debug = self.noc_block.get_risc_debug(risc_name="trisc0")
        base_address = self.noc_block.dest_start_address.private_address

        with risc_debug.ensure_halted():
            for i in range(num_tiles * TILE_SIZE):
                address = base_address + 4 * i
                rd_data = risc_debug.read_memory(address)
                data.append(self._unpack_value(rd_data, df))
```

This way we check `self.noc_block` only once, instead in a loop. Also, we avoid fetching `self.noc_block.dest_start_address.private_address` on every iteration, although that one is negligible. 
And most importantly, with is now outside the loop. You were re-checking is risc halted on every iteration, which is expensive. I assume it's safe to do once for the full read range.

adjordjevic-TT: Works, thanks!

fvranicTT (approved):

fvranicTT:
Overal, LGTM, but I am not the code owner.

ttexalens/debug_tensix.py
```diff
@@ -159,15 +162,66 @@ def write_tensix_register(self, register: str | RegisterDescription, value: int)
...
         """
-        device = self.context.devices[self.device_id]
-        noc_block = device.get_block(self.core_loc)
-        register_store = noc_block.get_register_store()
-        register_store.write_register(register, value)
+        self.register_store.write_register(register, value)
+
+    @staticmethod
+    def _validate_number_of_tiles(num_tiles: int | None) -> int:
+        max_num_tiles = 8  # maximum number of tiles in dest for 32 bit formats
```
tt-vjovanovic, line 169: This should come from definition in FunctionalWorker noc block...

ttexalens/debug_tensix.py
```diff
@@ -176,7 +230,7 @@ def read_regfile_data(
                 regfile (int | str | REGFILE): Register file to dump (0: SRCA, 1: SRCB, 2: DSTACC).
 
         Returns:
-                bytearray: 64x32 bytes of register file data (64 rows, 32 bytes per row).
+                list[int | float]: 64x32 bytes of register file data (64 rows, 32 bytes per row). Also returns num_tiles * TILE_SIZE of unpacked data if using direct dest reading.
```
tt-vjovanovic, line 234: Update comment, it is no longer `64x32 bytes`...

tt-vjovanovic (approved):

ttexalens/hardware/blackhole/functional_worker_block.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, location: OnChipCoordinate):
         self.l1 = MemoryBlock(
             size=1536 * 1024, address=DeviceAddress(private_address=0x00000000, noc_address=0x00000000)
         )
+        self.dest_start_address: DeviceAddress = DeviceAddress(private_address=0xFFBD8000)
```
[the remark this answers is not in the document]
adjordjevic-TT: Changed.

Issue #532, opened 2025-08-01
NOC command should show all 4 command buffers in sorted register dump with address


miacim:
The current output shows only the first command buffer, and registers are not ordered by address. Update the command to:

- Include all 4 command buffers.
- In WH, each is offset by 0x400 (0xFFB20000, 0xFFB20400, 0xFFB20800, 0xFFB20C00). In BH, offset is 0x800 between buffers.
- Sort all registers by address, with the lowest address first (e.g., command buffer registers should appear first).
- Display register addresses in the output.
 
Only command buffer registers have per-buffer copies [NOC_REGS_START_ADDR, NOC_REGS_START_ADDR +0x48]

PR #541, opened 2025-08-05
#532 Extending noc command to show address

miacim:
Fix for #532 

- added noc control registers for command buffer 1, 2 and 3
- added address column to output of noc command
- changed rich table to have configurable number of columns


ttexalens/hardware/blackhole/niu_registers.py
```diff
@@ -79,23 +79,78 @@
...
-    "NOC_AT_LEN_BE_1": NocControlRegisterDescription(offset=0x24),
-    "NOC_AT_DATA": NocControlRegisterDescription(offset=0x28),
-    "NOC_BRCST_EXCLUEDE": NocControlRegisterDescription(offset=0x2C),
-    "NOC_L1_ACC_AT_INSTRN": NocControlRegisterDescription(offset=0x30),
-    "NOC_SEC_CTRL": NocControlRegisterDescription(offset=0x34),
-    "NOC_CMD_CTRL": NocControlRegisterDescription(offset=0x40),
-    "NOC_NODE_ID": NocControlRegisterDescription(offset=0x44),
-    "NOC_ENDPOINT_ID": NocControlRegisterDescription(offset=0x48),
+    # Command buffer 0 registers (base offset 0x0)
+    "NOC_TARG_ADDR_LO_CB0": NocControlRegisterDescription(offset=0x0),
```
tt-vjovanovic, line 83: We need to figure out what makes more sense. We made registers with idea that they are at most 4 bytes wide (single read). But some buffers are bigger and maybe it makes sense to add different classes that will solve this problem...

tt-vjovanovic (approved):

ttexalens/hardware/blackhole/niu_registers.py
miacim: I do not have clear opinion here. Created #542

Issue #569, opened 2025-08-28
Build system improvements


iklikovacTT:
Created this issue from Vuk's comment on my code coverage PR while I was editing the build system.

Build improvement would be to have different directories for debug and release builds (e.g. build_debug, build_release) and build directory should be symbolic link to last chosen build.
Also, elfs should be built in different directory (like build_elfs) and added to build_* directory as symlink...
That way most of the tests/code/scripts/make/usage won't care about where build is and just use build directory as they used to :)

PR #571, opened 2025-08-29
Change build system to have debug and release builds

iklikovacTT:
#569 
Created `make debug` and `make release` options. They create `build_debug`/`build_release` and symlink `build` to that directory. Also created `build_elfs` and made `build/riscv-src` symlink to that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ core
 *.pkl
 
 .cpmcache
-/build
+/build*
 
 __pycache__
 .venv
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,18 @@ set(UMD_HOME "${PROJECT_SOURCE_DIR}/third_party/umd")
 # Third party dependencies
 add_subdirectory(${PROJECT_SOURCE_DIR}/third_party/umd)
 
+# Handle symlinks: build should point to build_debug or build_release,
+# and build_*/riscv-src should point to build_elfs
+set(BUILD_SYMLINK "${CMAKE_SOURCE_DIR}/build")
+set(ELF_SYMLINK "${CMAKE_BINARY_DIR}/riscv-src")
+file(MAKE_DIRECTORY "build_elfs")
+
+# Refresh the symlinks
+file(REMOVE "${BUILD_SYMLINK}")
+file(CREATE_LINK "${CMAKE_BINARY_DIR}" "${BUILD_SYMLINK}" SYMBOLIC)
+file(REMOVE "${ELF_SYMLINK}")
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -8,7 +8,7 @@ core
 *.pkl
 
 .cpmcache
-/build
+/build*
```
tt-vjovanovic, line 11: maybe better to be explicit:
```
/build
/build_debug
/build_release
/build_elfs
```
or whatever are the real names :)

riscv-src/CMakeLists.txt
```diff
@@ -135,11 +134,19 @@ function(create_riscv_target ARCH APP CORE BUILD_TYPE)
     add_dependencies("${ARCH}_${APP}_${CORE}_${BUILD_TYPE}_target" "${ARCH}_${APP}_${CORE}_${BUILD_TYPE}_dump_dis")
 endfunction()
 
+# On Debug builds, build every ELF variant (for tests etc.)
```
tt-vjovanovic, line 137: elf build should be "separate" but still triggered as other builds.
Separate means we want to build it once and for all flavors. Build once means if we do `make clean`, `make debug`, that will build elfs. Then if we run `make release` elfs are already built, so no need to do anything for ninja.

CMakeLists.txt
```diff
@@ -72,4 +84,4 @@ include(${PROJECT_SOURCE_DIR}/cmake/pybind.cmake)
 add_subdirectory(${PROJECT_SOURCE_DIR}/ttexalens)
 add_subdirectory(${PROJECT_SOURCE_DIR}/utils)
 add_subdirectory(${PROJECT_SOURCE_DIR}/test)
-add_subdirectory(${PROJECT_SOURCE_DIR}/riscv-src)
+add_subdirectory(${PROJECT_SOURCE_DIR}/riscv-src ${CMAKE_BINARY_DIR}/riscv-build)
```
tt-vjovanovic, line 87: What does this do?

riscv-src/CMakeLists.txt
iklikovacTT: I see, setting it up now.

.github/workflows/build-ttexalens.yml
```diff
@@ -45,15 +45,21 @@ jobs:
...
         run: |
-          make build
+          make debug
+          make release
       - name: Run mypy
         run: |
           make mypy
       - name: Build wheel
         run: |
+          make release
```
tt-vjovanovic, line 55: Remove this line.

tt-vjovanovic (approved):

miacim (approved):

PR #609, opened 2025-10-07
Upgrading CLI app state

tt-vjovanovic:
We can now start server even after we start CLI app.
We can switch which noc we are using as default communication with device using `go -n 0|1`.
We can start simulator and then start exalens server if needed.
Fixed small bug with running server in the background.

### DIFF
--- a/ttexalens/cli.py
+++ b/ttexalens/cli.py
@@ -4,8 +4,8 @@
 # SPDX-License-Identifier: Apache-2.0
 """
 Usage:
-  tt-exalens [--commands=<cmds>] [--start-gdb=<gdb_port>] [--devices=<devices>] [--verbosity=<verbosity>] [--test] [--jtag] [--use-noc1]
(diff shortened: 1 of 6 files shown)


ttexalens/cli.py
```diff
@@ -384,28 +401,30 @@ def main():
...
+            ttexalens_server = tt_exalens_server.start_server(port=int(args["--port"]), communicator=communicator)
+
             util.INFO("The debug server is running in the background.")
             util.INFO("To stop the server, use the command: touch exit.server")
-            os.remove("exit.server")
+            if os.path.exists("exit.server"):
+                os.remove("exit.server")
             try:
-                while os.path.isfile("exit.server"):
+                while not os.path.isfile("exit.server"):
```
adjordjevic-TT, line 418: Maybe add a comment why are we doing this

adjordjevic-TT (approved):

PR #630, opened 2025-10-15
Refactored exalens to use JTAG from UMD.

uvelimirovicTT:
- This PR tries to unify exalens API for both JTAG and PCIe protocols to use the same functions without relying on different device implementations.
- Completely removed jtag_implementation.
- Changed method names in ttexalens_implementation.
- Complete JTAG implementation relies now on passing the JTAG flag to cluster creation in open_implementation.

### DIFF
--- a/ttexalens/tt_exalens_lib.py
+++ b/ttexalens/tt_exalens_lib.py
@@ -91,9 +91,7 @@ def read_word_from_device(
     noc_id = check_noc_id(noc_id, context)
 
     noc_loc = context.convert_loc_to_umd(coordinate)
-    word = context.server_ifc.read32(
-        noc_id, coordinate.device_id, noc_loc[0], noc_loc[1], addr
-    )
+    word = context.server_ifc.read32(noc_id, coordinate.device_id, noc_loc[0], noc_loc[1], addr)
     return word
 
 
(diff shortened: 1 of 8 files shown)


ttexalens/tt_exalens_lib.py
```diff
@@ -175,15 +163,10 @@ def read_from_device(
...
-            coordinate, addr, coordinate.device_id, num_bytes // 4 + (num_bytes % 4 > 0), context, noc_id
-        )
-        return struct.pack(f"{len(int_array)}I", *int_array)[:num_bytes]
-
-    noc_loc = context.convert_loc_to_umd(coordinate)
-    return context.server_ifc.pci_read(noc_id, coordinate.device_id, noc_loc[0], noc_loc[1], addr, num_bytes)
+    
+    return context.server_ifc.read(
+        noc_id, coordinate.device_id, *context.convert_loc_to_umd(coordinate), addr, num_bytes
+    )
```
tt-vjovanovic: You need to update to latest code, `make mypy` could fail for you. See how removed code looks like:
```
noc_loc = context.convert_loc_to_umd(coordinate)
return context.server_ifc.pci_read(noc_id, coordinate.device_id, noc_loc[0], noc_loc[1], addr, num_bytes)
```

PR #677, opened 2025-11-04
Exposing UMD warm reset

### LINKED ISSUE #674: Expose warm reset from UMD
We want to test accessing remote chip when default active eth is hung. Hanging active eth breaks it therefore we need to reset chip to be able to start exalens again. Due to this we need to expose warm reset functionality from UMD and use it in tear down of this test suite.

adjordjevic-TT:
Closes #674 

Exposing warm reset functionality from UMD. To ensure reset is completed we need to call `init_exalens()` afterwards. Since we are not exposing this functionality to users and are only using it in our tests this should be fine for now.

### DIFF
--- a/riscv-src/globals_test.cc
+++ b/riscv-src/globals_test.cc
@@ -36,7 +36,35 @@ struct go_msg_t {
     uint32_t test2;
 } __attribute__((packed));
 
-struct GlobalStruct {
+struct BaseStruct {
+    uint8_t base_field1;
+    uint16_t base_field2;
+    union {
+        uint32_t packed;
+        struct {
+            uint8_t v1;
+            uint8_t v2;
+            uint8_t v3;
+            uint8_t v4;
+        };
+    };
+};
+
+struct BaseStruct2 {
+    uint8_t bs2_base_field1;
+    uint16_t bs2_base_field2;
+    union {
+        uint32_t bs2_packed;
+        struct {
+            uint8_t bs2_v1;
+            uint8_t bs2_v2;
+            uint8_t bs2_v3;
+            uint8_t bs2_v4;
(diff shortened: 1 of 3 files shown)


ttexalens/pybind/src/umd_implementation.cpp
```diff
@@ -252,4 +254,23 @@ std::optional<std::tuple<uint64_t, uint64_t, uint64_t>> umd_implementation::get_
...
+    bool is_4u_galaxy_configuration = cluster != nullptr && cluster->get_target_remote_device_ids().size() > 0 &&
+                                      cluster->get_cluster_description()->get_board_type(
+                                          *cluster->get_target_remote_device_ids().begin()) == BoardType::GALAXY;
+    bool is_6u_galaxy_configuration = cluster->get_target_device_ids().size() > 0 &&
+                                      cluster->get_cluster_description()->get_board_type(0) == BoardType::UBB;
+    return is_4u_galaxy_configuration || is_6u_galaxy_configuration;
+}
+
+void umd_implementation::warm_reset() {
+    if (_is_galaxy_configuration(cluster)) {
```
miacim: Why do we need this check? Shouldn't UMD handle this?

adjordjevic-TT: `UMD` purposefully chose to have reset functions because `BoardType` is not available in all scenarios therefore including check for galaxy would break reset sometimes. There is also runtime boolean variable available in `metal` which tells us if configuration is galaxy or not so it can be used to determine which reset we need to call. Knowing all this it would be best to have function argument `is_galaxy_configuration` which user can provide or even it could be read via inspector from `metal` runtime.

Issue #651, opened 2025-10-21
Tensix blackhole debug bus signals


vluzaninTT:
- Rename the debug bus signals for Blackhole so that all signals shared between Wormhole and Blackhole have the same names.
 - Also, it would be good to review the signal names on Wormhole and see if they can be renamed to make them more intuitive to use.

Relevant documentation:
 - `https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/DebugDaisychain.md`
 - `https://tenstorrent.atlassian.net/wiki/spaces/LLK/pages/1063321786/Debug+bus+access+and+map`

PR #696, opened 2025-11-12
Rename and add grouping for debug bus signals on blackhole

vluzaninTT:
Related issue: #651

- Rename the debug bus signals for Blackhole so that all signals shared between Wormhole and Blackhole have the same names.
 - Add debug bus signal grouping on Blackhole, as there is on Wormhole
 - Enable debug bus tests on Blackhole

Relevant documentation:
 - `https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/DebugDaisychain.md`
 - `https://tenstorrent.atlassian.net/wiki/spaces/LLK/pages/1063321786/Debug+bus+access+and+map`


miacim:
It makes sense to keep, not delete, old names and add new ones. Not sure if old names are already used by LLK team.

Issue #302, opened 2025-03-26
Refactor all CLI commands to use unified Rich formatting API


miacim:
### Description:
A new Rich-based formatting API (rich_formatters.py) has been introduced to standardize and improve the readability of CLI command outputs. This issue tracks the task of migrating all existing CLI commands to utilize this unified formatting mechanism, ensuring consistency across the entire toolset.

### Scope of Work:
- Update each CLI command module to replace any custom or inconsistent printing logic with calls to the new Rich formatter (formatter instance from rich_formatters.py).
- Verify that outputs from all commands align with the established styles (tables, headers, grouped data) and are visually consistent.
- Remove obsolete formatting code and dependencies after migration.

PR #703, opened 2025-11-14
Dumping tensix gpr

adjordjevic-TT:
Added tensix general purpose registers to `functional_worker_registers` `register_map`.
Renamed `dump-config-reg` to `dump-tensix-reg` and extended it command to also print these registers in 3 tables (one for each thread). By default command prints just registers named in `llk` and there is verbose mode (`-v`) for printing all registers.
Named them same as `llk` did here for [wh](https://github.com/tenstorrent/tt-llk/blob/main/tt_llk_wormhole/common/inc/ckernel_gpr_map.h) and [bh](https://github.com/tenstorrent/tt-llk/blob/main/tt_llk_blackhole/common/inc/ckernel_gpr_map.h). Those that are not named in `llk` I named them as `GPR_<index>_<thread_id>`.
Added few of them in `test_lib.py` where we test write/read register.

### DIFF
--- a/ttexalens/hardware/blackhole/functional_worker_registers.py
+++ b/ttexalens/hardware/blackhole/functional_worker_registers.py
@@ -808,15 +808,11 @@ def get_pack_strides() -> list[dict[str, str]]:
 def get_general_purpose_registers() -> list[dict[str, str]]:
     register_mapping_by_thread: list[dict[str, str]] = [dict() for _ in range(3)]
     for register_name in register_map.keys():
-        if not isinstance(register_map[register_name], TensixGeneralPurposeRegisterDescription):
-            continue
-        if register_name.endswith("_T0") or register_name.endswith("_T1") or register_name.endswith("_T2"):
-            thread_id = int(register_name[-1])  # Last character is the thread id
+        register_desc = register_map[register_name]
+        if isinstance(register_desc, TensixGeneralPurposeRegisterDescription):
+            assert isinstance(register_desc, TensixGeneralPurposeRegisterDescription)
+            thread_id = register_desc.thread_id
             register_mapping_by_thread[thread_id][register_name[:-3].lower()] = register_name
-        else:
-            thread_id = int(register_name[5])  # Fifth character is the thread id in this case
-            index = int(register_name[7:])  # Index is the number after the underscore
-            register_mapping_by_thread[thread_id][f"ID = {index}"] = register_name
 
     return register_mapping_by_thread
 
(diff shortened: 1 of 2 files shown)


ttexalens/hardware/wormhole/functional_worker_registers.py
```diff
@@ -819,6 +1023,22 @@ def get_pack_strides() -> list[dict[str, str]]:
...
+            register_mapping_by_thread[thread_id][register_name[:-3].lower()] = register_name
+        else:
+            thread_id = int(register_name[5])  # Fifth character is the thread id in this case
+            index = int(register_name[7:])  # Index is the number after the underscore
+            register_mapping_by_thread[thread_id][f"ID = {index}"] = register_name
+
+    return register_mapping_by_thread
+
+
 configuration_registers_descriptions = TensixConfigurationRegistersDescription(
```
tt-vjovanovic: Maybe we should rename `TensixConfigurationRegistersDescription`, into `TensixRegistersDescription`

ttexalens/hardware/blackhole/functional_worker_registers.py
```diff
@@ -595,6 +799,22 @@ def get_pack_strides() -> list[dict[str, str]]:
...
+        if not isinstance(register_map[register_name], GeneralPurposeRegisterDescription):
+            continue
+        if register_name.endswith("_T0") or register_name.endswith("_T1") or register_name.endswith("_T2"):
+            thread_id = int(register_name[-1])  # Last character is the thread id
+            register_mapping_by_thread[thread_id][register_name[:-3].lower()] = register_name
+        else:
+            thread_id = int(register_name[5])  # Fifth character is the thread id in this case
+            index = int(register_name[7:])  # Index is the number after the underscore
+            register_mapping_by_thread[thread_id][f"ID = {index}"] = register_name
+
```
tt-vjovanovic: Might be safer to do this with regex?
First if is leaving only 192 entries that you would run regex on. Create regex only once and it won't be noticeable on performance impact.

tt-vjovanovic (approved):

ttexalens/cli_commands/dump-tensix-reg.py
```diff
@@ -31,11 +35,12 @@
     "command_option_names": ["--device", "--loc"],
 }
 
-import copy
+
+import tabulate
```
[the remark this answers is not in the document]
adjordjevic-TT: Since gpr is part of `dump-tensix-reg` command we would need to update all of tables in that command that use API from `util` so we would essentially need to resolve whole #302 so I would keep that for another PR.

Issue #747, opened 2025-12-08
TT-triage causes host to reboot on dual 6u galaxy


adjordjevic-TT:
Model team encountered issue that running `tt-triage`, after running their workload, on dual 6u causes host to reboot.

We suspect the issue is we somehow kill `PCIe` core or hang `NOC`, we do not detect it, and keep trying to read from/write to device through that `PCIe` which crashes host. Currently, we try to detect this with timeout which if exceeded throws exception. Problem is that times for read/write after `PCIe` is broken or `NOC` is hung are not consistent across board types so it is difficult to set single value that will work everywhere for reads and writes of all sizes.

This is current plan for solving this:
1) If we always keep size at 4 bytes we could set timeout to very small value (which should catch `PCIe`/`NOC` problem everywhere). This will also solve issue where large read/write could break `PCIe`/`NOC` and continue reading/writing for long enough to crash host. We should check how much this would lower performance and then discuss if this should be default behavior.
2) We should test how much time valid reads/writes take across all board types, also we should test how much time they take after breaking `PCIe`/`NOC`. This should give us enough intel to set default timeout value.
3) Check if we break one device if we can access other chips on same board through their `PCIe`.
4) We should check if there is information about `PCIe` core health or `NOC` state that we can access. If this is the case, it would remove need of timeouts.

PR #752, opened 2025-12-08
Implemented 4 byte read/write mode

### LINKED ISSUE #748: Test single word read/write influence on tt-triage performance
Due to [issue](https://github.com/tenstorrent/tt-exalens/issues/747) we consider switching to only single word reads in `tt-triage`.
We already tested this and we lowered speed from `4MB/s` to `600-700KB/s` but since `tt-triage` does not have many big reads we should see how much this actually influences its performance.

### LINKED ISSUE #758: Make read/write timeout configurable through env variable
We should be able to change timeout to account for new scenarios without needing to rebuild `tt-exalens`. We can do this with env variable which we would inspect only when we are initializing context.

adjordjevic-TT:
Closes #748
Closes #758 

Introduced 4-byte mode which forces all reads and writes to access memory word by word to prevent issues with host (#747). When creating context we default to using 4-byte therefore triage will automatically use this mode. Here are results of testing triage with and without changes:
[triage_time_4B.txt](https://github.com/user-attachments/files/24037837/triage_time_4B.txt)
[triage_time.txt](https://github.com/user-attachments/files/24037851/triage_time.txt)
Total time per device has increased from `3.66s` to `4.74s` which is acceptable.

Application can also use this mode and it is turned on by default and `go` command is extended to be able to switch between modes. We also print if we are using 4B mode in application prompt.

Since we are now only accessing memory word by word we were able to lower timeouts to 2 ms. Reenabled write timeout but only for 4B chunks and only throwing exception if 5 consecutive writes timeout.

Added 3 env variables:
1) `TT_EXALENS_READ_TIMEOUT_MS` used for setting timeout for `read_from_device_reg` function
2) `TT_EXALENS_WRITE_TIMEOUT_MS` used for setting timeout for `write_to_device_reg` function
3) `TT_EXALENS_NUM_OF_CONSECUTIVE_TIMEOUTS` used for setting maximum number of consecutive writes that timeout before we throw exception.

---

> [!NOTE]
> Add default-on 4B (word-by-word) device read/write mode across stack with CLI toggle/prompt, API plumbing, UMD chunking, and docs/tests updates, plus tighter timeout handling.
> 
> - **Core/UMD**:
>   - Add `use_4B_mode` to read/write paths; implement 4B chunking in unaligned helpers.
>   - Lower device register timeouts to 2ms; re-enable write timeout with exception only after 5 consecutive timeouts; include elapsed duration in timeout error.
> - **Pybind Interface**:
>   - Extend `read`/`write` bindings and implementation interfaces to accept `use_4B_mode`.
> - **Library/API (`tt_exalens_lib`, `tt_exalens_init`, `context`)**:
>   - Propagate `use_4B_mode` through context; default to True; add parameter to read/write helpers (`read_words_from_device`, `read_from_device`, `write_*`).
> - **CLI**:
>   - Add `--disable-4B-mode` global flag; show `[4B MODE]` in prompt when enabled.
>   - Extend `go` command with `-m <mode>` to toggle 4B mode; keep `-n` for NOC selection.
> - **Tests**:
>   - Update prompt regex to allow optional `[4B MODE]` indicator.
> - **Docs**:
>   - Update app and lib docs to reflect 4B mode defaults, new CLI/options, and adjusted outputs; rename/standardize command section headers (e.g., `burst-read-xy / brxy`, `debug-bus / dbus`, `dump-coverage / cov`, `dump-gpr / gpr`, `tensix-reg / reg`, `write-xy / wxy`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 34062ac73cb0861e8d0e25325af4f649147d1ea0. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/riscv-src/callstack.cc
+++ b/riscv-src/callstack.cc
@@ -13,7 +13,8 @@ void halt() {
 }
 
 int f1(int a) {
-    if (a <= 1) {
+    constexpr int recursion_end = 1;
+    if (a <= recursion_end) {
         halt();
         return a;
     } else
--- a/test/ttexalens/unit_tests/test_lib.py
+++ b/test/ttexalens/unit_tests/test_lib.py
@@ -957,7 +957,7 @@ class TestCallStack(unittest.TestCase):
                 self.assertEqual(entry1.pc, entry2.pc, "Addresses do not match")
 
     CALLSTACK_ELFS = ["callstack.debug", "callstack.release", "callstack.coverage"]
-    RECURSION_COUNT = [1, 10, 50]
+    RECURSION_COUNT = [1, 10, 40]
 
     @parameterized.expand(itertools.product(CALLSTACK_ELFS, RECURSION_COUNT))
     def test_callstack_with_parsing(self, elf_name, recursion_count):
@@ -1033,24 +1033,6 @@ class TestCallStack(unittest.TestCase):
         callstack: list[CallstackEntry] = lib.top_callstack(pc, elf_path, None, self.context)
         self.assertEqual(len(callstack), 1)
         self.assertEqual(callstack[0].function_name, "halt")
-
-    @parameterized.expand(itertools.product(CALLSTACK_ELFS, RECURSION_COUNT))
-    def test_callstack_optimized(self, elf_name: str, recursion_count: int):
-        lib.write_words_to_device(self.location, 0x64000, recursion_count)
-        elf_path = self.get_elf_path(elf_name)
-        self.loader.run_elf(elf_path)
-        callstack: list[CallstackEntry] = lib.callstack(self.location, elf_path, None, self.risc_name, None, 100, True)
-
-        self.assertEqual(len(callstack), recursion_count + 3)
-        self.assertEqual(callstack[0].function_name, "halt")
-        for i in range(1, recursion_count):
-            self.assertEqual(callstack[i].function_name, "f1")
-        self.assertEqual(callstack[recursion_count + 1].function_name, "recurse")
-        self.assertEqual(callstack[recursion_count + 2].function_name, "main")
-        gdb_callstack: list[CallstackEntry] = get_gdb_callstack(
-            self.location, self.risc_name, [elf_path], [None], self.gdb_server
-        )
-        self.compare_callstacks(callstack, gdb_callstack)
 
     @parameterized.expand([(1, 1)])
     def test_top_callstack_optimized(self, recursion_count: int, expected_f1_on_callstack_count: int):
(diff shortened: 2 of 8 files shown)


docs/ttexalens-lib-docs.md
```diff
@@ -58,12 +59,11 @@ Interfacing device is done remotely through TTExaLens client.
...
 - `port` *(int)*: Port number of the TTExaLens server interface. Default is 5555.
+- `use_4B_mode` *(bool)*: Whether to use 4B mode for communication with the device. Default is True.
+- `Returns`
+- `Context`: TTExaLens context object.
 
 
-### Returns
-
- *(Context)*: TTExaLens context object.
-
```
tt-vjovanovic: This looks wrong

ttexalens/cli.py
```diff
@@ -26,6 +26,7 @@
   --test                          Exits with non-zero exit code on any exception.
   --jtag                          Initialize JTAG interface.
   --use-noc1                      Initialize with NOC1 and use NOC1 for communication with the device.
+  --disable-4B-mode               Disable4B mode for communication with the device.
```
tt-vjovanovic:
```suggestion
  --disable-4B-mode               Disable 4-byte mode for communication with the device.
```

ttexalens/tt_exalens_init.py
```diff
@@ -38,27 +40,28 @@ def init_ttexalens(
...
+    use_4B_mode: bool = True,
 ) -> Context:
     """Initializes TTExaLens internals by creating the device interface and TTExaLens context.
     Interfacing device is done remotely through TTExaLens client.
 
     Args:
             ip_address (str): IP address of the TTExaLens server. Default is 'localhost'.
             port (int): Port number of the TTExaLens server interface. Default is 5555.
-
+            use_4B_mode (bool): Whether to use 4B mode for communication with the device. Default is True.
```
tt-vjovanovic: You need empty line after to split `Args:` and `Returns:` sections

tt-vjovanovic:
```suggestion
            use_4B_mode (bool): Whether to use 4-byte mode for communication with the device. Default is True.
```

tt-vjovanovic (approved):

docs/ttexalens-lib-docs.md
adjordjevic-TT: Fixed

ttexalens/pybind/src/umd_implementation.cpp
```diff
@@ -18,15 +18,34 @@
 
 namespace tt::exalens {
 
+std::chrono::microseconds READ_TIMEOUT_US = []() -> std::chrono::microseconds {
```
tt-vjovanovic: You don't need to store value in `std::chrono::microseconds`, you can use `std::chrono::milliseconds`. There are automatic conversions happening when using comparison operators and assignment operators...

tt-vjovanovic: When using `std::chrono::*` structures, you don't need to use units in name as it is in type.

PR #760, opened 2025-12-10
Enable noc-accessible addresses on RiscDebugMemoryAccess

### LINKED ISSUE #745: Fix RiscDebugMemoryAccess so it can read NCRISC private memory
NCRISC private memory is available on NOC.
In `RiscDebugMemoryAccess` we check if `risc_debug` supports debugging (which is not true for NCRISC), but we can still read its private memory if it is not in reset on blackhole.

onenezicTT:
Closes #745

NCRISC private memory is available on NOC.
In RiscDebugMemoryAccess we check if risc_debug supports debugging (which is not true for NCRISC), but we can still read its private memory if it is not in reset on blackhole.

---

> [!NOTE]
> Allow `RiscDebugMemoryAccess` to access memory via NOC when debugging isn’t available, with access gating and reset checks.
> 
> - **Elf/Debug Access**:
>   - `RiscDebugMemoryAccess`:
>     - Initialize `private_memory` directly and introduce `can_access` (true if private memory exists and `can_debug()` or `exposed_through_noc()` is true).
>     - `read`/`write`:
>       - Early-exit when `can_access` is false; return early if RISC is in reset for non-private paths.
>       - Gate private-memory operations on `can_debug()`; otherwise fall back to device I/O.
> - **Hardware**:
>   - `MemoryBlock`: add `exposed_through_noc()` helper to indicate NOC exposure.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8566f4e137870b44d4d474a69f7f4a6a9e9448bf. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/ttexalens/elf/variable.py
+++ b/ttexalens/elf/variable.py
@@ -63,14 +63,16 @@ class L1MemoryAccess(MemoryAccess):
 class RiscDebugMemoryAccess(MemoryAccess):
     def __init__(self, risc_debug: RiscDebug):
         self._risc_debug = risc_debug
-        private_memory = risc_debug.get_data_private_memory()
-        if private_memory is None or not risc_debug.can_debug():
-            self.private_memory = MemoryBlock(0, DeviceAddress(-1))
-        else:
-            self.private_memory = private_memory
+        self.private_memory = risc_debug.get_data_private_memory()
+        self.can_access = (self.private_memory is not None) and (
+            risc_debug.can_debug() or self.private_memory.exposed_through_noc()
+        )
 
     def read(self, address: int, size_bytes: int) -> bytes:
-        if self.private_memory.contains_private_address(address):
+        if not self.can_access:
+            return 0
+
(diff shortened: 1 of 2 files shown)


ttexalens/elf/variable.py
```diff
@@ -63,14 +63,16 @@ def write(self, address: int, data: bytes) -> None:
...
-            self.private_memory = private_memory
+        self.private_memory = risc_debug.get_data_private_memory()
+        self.can_access = (self.private_memory is not None) and (
+            risc_debug.can_debug() or self.private_memory.exposed_through_noc()
+        )
 
     def read(self, address: int, size_bytes: int) -> bytes:
-        if self.private_memory.contains_private_address(address):
+        if not self.can_access:
+            return 0
```
onenezicTT: Usually should we raise something here or just swallow it? Same question for below..

PR #768, opened 2025-12-11
Enhancing mypy configuration

tt-vjovanovic:
Changing any of the `Relaxing rules for the whole project` causes build break.

Making checks more strict and fixing few of the new strict rules.

---

> [!NOTE]
> Tightens mypy config and adds/adjusts type hints, casts, and minor logic tweaks across CLI, ELF, hardware debug, and utils to satisfy stricter typing.
> 
> - **Config**:
>   - **mypy**: Enable stricter checks (e.g., disallow unimported Any, untyped decorators/globals, strict option tweaks), use `exclude_gitignore`, narrow excludes.
> - **Typing/Refactors (broad)**:
>   - Add explicit return/param types, generics, TYPE_CHECKING imports, and safe casts across `ttexalens/*` (CLI commands, `context`, `device`, `coordinate`, `elf/*`, `gdb`, `hardware/*`, `util`, formatters, pack/unpack, memory map, init/ifc/lib).
> - **Small functional adjustments**:
>   - Normalize watchpoint state return via wrapper in `BabyRiscDebug.read_watchpoints_state`.
>   - `Context.convert_loc_to_umd` now returns internal `location._noc0_coord`.
>   - CLI `run(...)` signatures (e.g., `noc`, `dump_coverage`) return `list[dict[str,str]]`.
>   - Ensure numeric/string conversions (e.g., `int(...)`, `float(...)`, `str(...)`) in ELF parsing, GDB port selection, register/memory reads.
>   - Minor frame/CFA handling tweaks in `elf/frame.py` and callstack utilities.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 432978bb15ec639088df54963b77a8d8cbb6aacf. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -35,16 +35,12 @@ disallow_any_expr = False
 disallow_any_decorated = False
 disallow_any_explicit = False
 disallow_any_generics = False
-
 disallow_untyped_calls = False
 disallow_untyped_defs = False
 disallow_incomplete_defs = False
-
 extra_checks = False
 implicit_reexport = True
-
 strict_equality = False
 strict_bytes = False
 strict = False
-
 warn_unreachable = False


ttexalens/elf/die.py
```diff
@@ -125,7 +126,7 @@ def path(self) -> str | None:
             dwarf_die = self.dwarf_die.get_DIE_from_attribute("DW_AT_specification")
             die = self.cu.dwarf.get_die(dwarf_die)
             if die is not None:
-                return die.path
+                return die.path  # type: ignore
```
onenezicTT: why does this complain and you needed an ignore?

ttexalens/hardware/baby_risc_debug.py
```diff
@@ -183,7 +183,7 @@ def get_register_name(reg_index_or_name):
 @dataclass
 class BabyRiscDebugStatus(RiscDebugStatus):
     @staticmethod
-    def from_register(value: int, max_watchpoints: int):
+    def from_register(value: int, max_watchpoints: int) -> "BabyRiscDebugStatus":
```
onenezicTT: i saw these quotation marks when retuning stuff from methods throughout the code, why do we use them?

ttexalens/hw/tensix/blackhole/blackhole_ops.py
```diff
@@ -1,6 +1,10 @@
 # SPDX-FileCopyrightText: © 2024 Tenstorrent AI ULC
 
 # SPDX-License-Identifier: Apache-2.0
+
+# Disabling mypy for this file
```
onenezicTT: why? :)

ttexalens/hardware/baby_risc_debug.py
tt-vjovanovic: Because type is still not defined.

ttexalens/elf/die.py
tt-vjovanovic: Because `mypy` doesn't work well with `@cached_property`.

PR #773, opened 2025-12-15
Expose reading N bytes from risc_debug

### LINKED ISSUE #745: Fix RiscDebugMemoryAccess so it can read NCRISC private memory

onenezicTT:
This PR adds two methods `read_memory_bytes` and `write_memory_bytes` to `baby_risc_debug` and efectively moves the additional reading logic out of `MemoryAccess` down, while solving the issue of `MemoryAccess` not reading data from ncrisc as it only relied on debug hardware. 

These two methods make alignments and fixups if the address is not aligned so the caller does need to worry about that which in turn fixes some stuff left in code from before (like when we overwrite words when we write unaligned on Blackhole).

Closes #745

### DIFF
--- a/test/ttexalens/unit_tests/test_risc_debug.py
+++ b/test/ttexalens/unit_tests/test_risc_debug.py
@@ -277,13 +277,54 @@ class TestDebugging(unittest.TestCase):
         # Verify third word is unchanged
         self.assertEqual(self.core_sim.read_data(addr + 8), 0x99AABBCC, "Third word should be unchanged")
 
-    def test_read_write_memory_bytes_unaligned(self):
+    @parameterized.expand(
+        [
+            # (offset, size, expected_read, write_data)
+            # Aligned cases
+            (0, 4, b"\x78\x56\x34\x12", b"\xAA\xBB\xCC\xDD"),
+            (0, 8, b"\x78\x56\x34\x12\xdd\xcc\xbb\xaa", b"\x11\x22\x33\x44\x55\x66\x77\x88"),
+            (4, 4, b"\xdd\xcc\xbb\xaa", b"\xEE\xFF\x00\x11"),
+            # Single byte at each offset
+            (0, 1, b"\x78", b"\xF0"),
+            (1, 1, b"\x56", b"\xF1"),
+            (2, 1, b"\x34", b"\xF2"),
+            (3, 1, b"\x12", b"\xF3"),
+            # Two bytes at each offset
+            (0, 2, b"\x78\x56", b"\xE0\xE1"),
+            (1, 2, b"\x56\x34", b"\xE2\xE3"),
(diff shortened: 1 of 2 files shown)


ttexalens/hardware/baby_risc_debug.py
```diff
@@ -712,13 +712,67 @@ def read_memory(self, address: int) -> int:
...
+        aligned_end = ((address + size_bytes + word_size - 1) // word_size) * word_size
+
+        words: list[int] = []
+        words_to_read = (aligned_end - aligned_start) // word_size
+        for offset in range(words_to_read):
+            new_addr = aligned_start + offset * word_size
+            word: int = self.debug_hardware.read_memory(new_addr)
+            words.append(word)
+
+        result = b"".join(word.to_bytes(word_size, byteorder="little") for word in words)
```
tt-vjovanovic: Maybe it is better to use `bytesarray` instead of two lists:

```suggestion
        result = bytearray()
        words_to_read = (aligned_end - aligned_start) // word_size
        for offset in range(words_to_read):
            new_addr = aligned_start + offset * word_size
            word: int = self.debug_hardware.read_memory(new_addr)
            result.extend(word.to_bytes(4, byteorder="little"))
```

ttexalens/hardware/blackhole/baby_risc_debug.py
```diff
@@ -24,79 +23,58 @@ def read_gpr(self, register_index: int) -> int:
...
+        data = value.to_bytes(4, byteorder="little")
+        self.write_memory_bytes(address, data)
+
+    def read_memory_bytes(self, address: int, size_bytes: int):
         if self.enable_asserts:
             self.assert_not_in_reset()
 
-        write_memory: Callable[[int, int], int]
-
+        read_memory: Callable[[int, int], bytes]
```
tt-vjovanovic: Now you don't need lambda. You can just return what was read on line 43 and line 48.

ttexalens/hardware/blackhole/baby_risc_debug.py
```diff
@@ -24,79 +23,58 @@ def read_gpr(self, register_index: int) -> int:
...
 
-    def read_memory(self, address: int):
+        return read_memory(address, size_bytes)
+
+    def write_memory_bytes(self, address: int, data: bytes):
         if self.enable_asserts:
             self.assert_not_in_reset()
 
-        read_memory: Callable[[int], int]
+        write_memory: Callable[[int, bytes], None]
```
tt-vjovanovic: Remove lambda and directly call `write_to_device` and `super().write_memory_bytes`

Issue #729, opened 2025-11-28
Confirm that the RWC debug bus signals on Blackhole are correct


vluzaninTT:
With the LLK team, verify that reading the RWC debug-bus signals on Blackhole is correct. The following LLK test (tt-llk/tests/sources/rwc_test.cpp) writes values into the RWC registers for all three threads.

```cpp
// SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
//
// SPDX-License-Identifier: Apache-2.0

#include <cstdint>

#include "ckernel.h"
#include "ckernel_instr_params.h"
#include "ckernel_ops.h"

#ifdef LLK_TRISC_UNPACK

using namespace ckernel;

void run_kernel()
{   
    // Zero all counters
    TTI_SETRWC(0, 0, 0, 0, 0, p_setrwc::SET_ABD);   //rwc_a=0, rwc_b=0, rwc_d=0

    TTI_SETRWC(0, 0, 0xf, 0x09, 0x03, p_setrwc::SET_ABD); // rwc_a=4, rwc_b=8, rwc_d=15
    asm volatile("ebreak");
}
#endif

#ifdef LLK_TRISC_MATH

using namespace ckernel;

void run_kernel()
{
    // Zero all counters
    TTI_SETRWC(0, 0, 0, 0, 0, p_setrwc::SET_ABD);   //rwc_a=0, rwc_b=0, rwc_d=0

    TTI_SETRWC(0, 0, 0xf, 0x08, 0x04, p_setrwc::SET_ABD); // rwc_a=4, rwc_b=8, rwc_d=15
    asm volatile("ebreak");

    // Maximum values (check bit widths)
    // TTI_SETRWC(0, 0, 0x3FF, 0x3F, 0x3F, p_setrwc::SET_ABD);  // rwc_a=63 (6-bit max), rwc_b=63, rwc_d=1023 (10-bit max)

    // Alternating bit patterns
    // TTI_SETRWC(0, 0, 0x2AA, 0x2A, 0x15, p_setrwc::SET_ABD); // rwc_a=0x15 (10101), rwc_b=0x2A (101010), rwc_d=0x2AA (1010101010)

    // Power of 2 values
    // TTI_SETRWC(0, 0, 0x200, 0x20, 0x10, p_setrwc::SET_ABD);
    // rwc_a=16, rwc_b=32, rwc_d=512

    // Set only srcA counter
[...]

PR #793, opened 2025-12-23
Extending dump-tensix-reg with rwc and adc making dump-tensix-state

adjordjevic-TT:
Extending `dump-tensix-reg` command with RWCs and ADCs read from debug bus and renaming it to `dump-tensix-state`.
Those groups are printed by default. If L1 address is provided we use atomic group reading, if not we use newly added unsafe group reading which reads group through debug hardware with 4 reads (4x32 bit =128 bit).
Similar to `TensixRegisterDescription` added `TensixDebugBusDescription` that currently stores group names for RWC and ADC. 

RWCs are currently not working on `blackhole` so we skip them on that arch. #729

### DIFF
--- a/ttexalens/cli_commands/dump-tensix-state.py
+++ b/ttexalens/cli_commands/dump-tensix-state.py
@@ -6,12 +6,12 @@ Usage:
   dump-tensix-state [ <group> ] [ -d <device> ] [ -l <loc> ] [ -v ] [ -t <thread-id> ] [ -a <l1-address> ]
 
 Options:
-  <group>           Tensix group to dump. Options: [all, alu, pack, unpack, gpr, rwc, adc]
+  <group>           Tensix state group to dump. Options: [all, alu, pack, unpack, gpr, rwc, adc]
   -t <thread-id>    Thread ID. Options: [0, 1, 2] Default: all (only for gpr group)
   -a <l1-address>   L1 address to save group data. Only used for RWC and ADC groups.
 
 Description:
-  Prints the tensix group of the given name, at the specified location and device.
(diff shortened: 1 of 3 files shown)


ttexalens/cli_commands/dump-tensix-state.py
```diff
@@ -177,3 +193,51 @@ def run(cmd_text, context, ui_state: UIState):
                         )
                     )
                 print(put_table_list_side_by_side(tables))
+
+            if group == "rwc" or group == "all":
```
miacim: Code duplication should be avoided.
Ideally, we should separate data retrieval from presentation. We can expose an API that gathers all required data, and have a thin layer responsible only for printing/formatting.
This would also allow the same API to be reused later from triage.

ttexalens/debug_bus_signal_store.py
```diff
@@ -462,6 +467,41 @@ def read_signal_group(
...
+                    add_signal_to_read(signal_desc)
+
+        return signals_to_read
+
+    def read_signal_group_unsafe(self, signal_group: str) -> SignalGroupSample:
+        signals_to_read = self._find_signals_to_read(signal_group)
+        data: int = 0  # 128-bit data
+        for signal_desc in signals_to_read:
+            data |= self._read_signal_data(signal_desc) << (WORD_SIZE_BITS * signal_desc.rd_sel)
+        return SignalGroupSample(data, self.signal_groups[signal_group])
```
tt-vjovanovic: Just read all 4 words into data. No need to look for all different `rd_sel` in the group itself.

adjordjevic-TT: You are right. I was overthinking it.

PR #808, opened 2025-12-29
Promote MemoryAccess to a first-class concept

onenezicTT:
Currently, `MemoryAccess` is a concept defined int elf's `variable.py` and used throughout the repo (and in some places in triage). The idea is for it to evolve to be **the** abstraction for reading and writing data to and from the memory and make those reads explicit.

The first step towards that is to promote it into a first-class concept that will be exposed outside and inside exalens. Then it can be extended with with different needs (like maybe a specialization called `NocMemoryAccess` that reads only from NoC always, or from PCI/arc).

The next step is going through potential places for applying this change and analyzing its requirements (like moving `register_store`'s)

### DIFF
--- a/ttexalens/memory_access.py
+++ b/ttexalens/memory_access.py
@@ -10,7 +10,7 @@ from ttexalens.hardware.memory_block import MemoryBlock
 
 if TYPE_CHECKING:
     from ttexalens.coordinate import OnChipCoordinate
-    from ttexalens.hardware.risc_debug import RiscDebug
+    from ttexalens.hardware.risc_debug import RiscDebug, RiscLocation
 
 
 class RestrictedMemoryAccessError(Exception):
@@ -19,7 +19,38 @@ class RestrictedMemoryAccessError(Exception):
     (e.g., outside L1 or data private memory when restricted_access for them is is enabled).
     """
 
-    pass
+    def __init__(self, access_start: int, access_end: int, location: OnChipCoordinate | RiscLocation):
+        """
+        Args:
+            access_start: Starting address of the attempted access
+            access_end: Ending address of the attempted access (inclusive)
+            location: Location of the RISC or on-chip memory where the access was attempted
+        """
+
+        from ttexalens.coordinate import OnChipCoordinate
+
+        self.access_start = access_start
+        self.access_end = access_end
+        if isinstance(location, OnChipCoordinate):
+            self.risc_name = None
+            self.neo_id = None
+            self.location = location
+        else:
+            self.risc_name = location.risc_name
+            self.neo_id = location.neo_id
+            self.location = location.location
+
+    def __str__(self) -> str:
+        """Generate error message lazily when the exception is converted to string."""
+
+        msg = f"Restricted access: Address range [0x{self.access_start:08x}, 0x{self.access_end:08x}] is outside allowed memory regions."
+        if self.risc_name:
+            if self.neo_id is not None:
+                return f"RISC '{self.risc_name}' (Neo ID {self.neo_id}) at {self.location.to_user_str()}, {msg}"
+            else:
+                return f"RISC '{self.risc_name}' at {self.location.to_user_str()}, {msg}"
+
+        return f"{self.location.to_user_str()}, {msg}"
 
 
 class MemoryAccess(ABC):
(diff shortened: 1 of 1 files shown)


ttexalens/elf/__init__.py
```diff
@@ -6,7 +6,8 @@
 from .die import ElfDie
 from .dwarf import ElfDwarf
 from .parsed import ParsedElfFile, ParsedElfFileWithOffset, read_elf
-from .variable import ElfVariable, MemoryAccess
+from .variable import ElfVariable
+from ttexalens.memory_access import MemoryAccess
```
tt-vjovanovic: If we moved `MemoryAccess` to different module, we should adjust changes. It is not part of `ttexalens.elf` any more.

ttexalens/elf_loader.py
```diff
@@ -81,15 +82,23 @@ def write_block_through_debug(self, address, data):
...
+        mem.write(address, data)
 
     def read_block_through_debug(self, address, byte_count):
         """
         Reads a block of data from a given address through the debug interface.
         """
-        with self.risc_debug.ensure_private_memory_access():
-            data = self.risc_debug.read_memory_bytes(address, byte_count)
+        mem = MemoryAccess.get(
+            self.risc_debug,
```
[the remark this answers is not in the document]
onenezicTT: Yeah.. and also there's a concern that we currently just instantiate and dispose every time we want to do a read or a write. This might be cheap now as `MemoryAccess` currently does not have much in it but it can potentially be a problem, or might already be one if we have calling patterns that call `risc_debug.read` a bunch of times

ttexalens/elf/frame.py
```diff
@@ -40,13 +40,20 @@ def read_register(self, register_index: int, cfa: int) -> int | None:
...
-                    private_data is not None and private_data.contains_private_address(address)
-                ):
-                    return self.risc_debug.read_memory(address)
-                return None
+                try:
+                    mem = MemoryAccess.get(
+                        self.risc_debug,
+                        ensure_halted_access=True,
+                        restricted_access=True,
+                    )
```
tt-vjovanovic: Store this `MemoryAccess` into `FrameDescription` class as a member instead of creating object over and over again.

ttexalens/elf/__init__.py
onenezicTT: Sure, I wanted to keep the backwards compatibility for triage and then 1) clean up imports there and then 2) remove this backcomp after, but it's simpler to just clean that up during the bump

ttexalens/memory_access.py
```diff
@@ -0,0 +1,129 @@
...
+        return RiscDebugMemoryAccess(
+            risc_debug, ensure_halted_access=ensure_halted_access, restricted_access=restricted_access
+        )
+
+    @staticmethod
+    def get_l1(location: OnChipCoordinate) -> "MemoryAccess":
+        return L1MemoryAccess(location)
+
+
+class L1MemoryAccess(MemoryAccess):
```
adjordjevic-TT: Can we add comment explaining what this class does?

ttexalens/memory_access.py
```diff
@@ -0,0 +1,129 @@
...
+                and data_private_memory.contains_private_address(address + size_bytes - 1)
+            )
+
+            if not inside_l1 and not inside_data_private_memory:
+                raise Exception(
+                    f"RiscDebugMemoryAccess restricted access: Address 0x{address:08x} is outside of L1 and Data Private Memory"
+                )
+
+
+class CachedReadMemoryAccess(MemoryAccess):
```
adjordjevic-TT: Comment

adjordjevic-TT (approved):

ttexalens/elf/frame.py
```diff
@@ -40,13 +41,14 @@ def read_register(self, register_index: int, cfa: int) -> int | None:
...
-                    private_data is not None and private_data.contains_private_address(address)
-                ):
-                    return self.risc_debug.read_memory(address)
-                return None
+                try:
+                    return self.mem_access.read_word(address)
+                except Exception as e:
+                    # If access was restricted (outside L1/data_private_memory), return None
+                    if "restricted access" in str(e):
+                        return None
```
tt-vjovanovic: Please create issue for tracking this.
We would like to create new exception type for restricted access so that it can be easily filtered in `except` block instead of catching all exceptions and searching through exception text.

tt-vjovanovic (approved):

tt-vjovanovic (approved):

PR #812, opened 2026-01-03
Refactoring IFC after move to UMD Python API

tt-vjovanovic:
Removal of lots of small unused code:
- `tt-lens` entry point in wheel
- `c++.natvis` - moved to UMD as we don't have C++ code on host
- `wanted_devices` were never used, but exposed as UMD cluster constructor had it - deprecated by `TT_VISIBLE_DEVICES`
- `LimitedContext` as we have only one context - `MetalContext` should be created in `tt-triage`
- `pci_read_tile` as it was not being used since `BudaBackend`
- `arch` property from `Context` - every device has its own `arch`

Refactoring how we use UMD API now instead of `server_ifc`, split into two: `umd_api`, `file_api`.
Two new classes for IFC:
- `UmdApi` - high level API unrelated to single device (device enumeration, reset, cluster descriptor, etc.)
- `UmdDevice` - device API - read/write, SOC descriptor, failure detection and retry for read/write, etc.

Expanding Pyro5 server to support sending UMD objects and exposing UMD interfaces.

Depends on https://github.com/tenstorrent/tt-umd/pull/1747.

### DIFF
--- a/ttexalens/server.py
+++ b/ttexalens/server.py
@@ -2,6 +2,7 @@
 
 # SPDX-License-Identifier: Apache-2.0
 from __future__ import annotations
+from dataclasses import dataclass
 import inspect
 import io
 import Pyro5.api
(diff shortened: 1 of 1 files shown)


test/ttexalens/unit_tests/test_lib.py
```diff
@@ -106,8 +104,6 @@ def test_write_read_bytes(self):
         data = b"abcd"
 
         ret = lib.write_to_device(location, address, data)
-        self.assertEqual(ret, len(data))
```
adjordjevic-TT: If we are not using return value of `write_to_device` we should not store it.

onenezicTT (approved):

PR #819, opened 2026-01-08
Protect reading/writing memory through GdbServer

### LINKED ISSUE #785: Protect reading/writing memory through GdbServer
Currently, getting callstack can break risc because stack walk can get outside of the stack. We need to limit what gdb can read/write on risc to be only L1 and private data memory.
Problem, understand how to return partial result to `x` and `m` GDB commands and how to fail `X` and `M` so that GDB understands that memory is not correct.

onenezicTT:
Closes #785 

This PR adds a restriction to gdb's reads and writes (`x/X/m/M`) to only L1 and DataPrivateMemory. We will not allow partial reads and especially not partial writes - gdb supports partial reads on some high level commands like `print` (you ask for X bytes, but return Y) which in return shows us a `$<incomplete type>`, at lower level it's probably an error that the higher level transforms into that incomplete type, sometimes it transforms and can potentially put is in some weird and undefined states with gdb. We have a new error code `E04` will signify this being a memory access error - if a lower level command like x/X/m/M tried a restricted read it will raise it, for a higher level command like `print` that error will be raised but transformed into `$<incomplete type>`.

Also added a bunch of tests that test the server against just `x/X/m/M` as well as just spawn our gdb client and test it as an integration test.

### DIFF
--- a/test/ttexalens/unit_tests/test_gdb_mem_access.py
+++ b/test/ttexalens/unit_tests/test_gdb_mem_access.py
@@ -114,7 +114,7 @@ class TestGdbMemAccessFromClient(unittest.TestCase):
         # Small delay to ensure the server is listening
         time.sleep(0.2)
 
-        cls.gdb_bin: str = get_gdb_client_path()
+        cls.gdb_bin = get_gdb_client_path()
 
     @classmethod
     def tearDownClass(cls) -> None:


test/ttexalens/unit_tests/test_gdb_mem_access.py
tt-vjovanovic: While it might be useful to have test that verifies current behavior of GdbServer and its communication, it will be better to test it through GDB client.
Test that you are introducing verifies that behavior is the one that we have, while it doesn't mean that GDB client works as expected with it...

PR #832, opened 2026-01-12
Add write verification option to ELF loading functions

tt-vjovanovic:
We verified writing elfs to memory by default in exalens. Adding option to disable it.

### DIFF
--- a/docs/bin/doc_utils.py
+++ b/docs/bin/doc_utils.py
@@ -45,7 +45,12 @@ class ElementPPrinter:
         return result
 
     def print_arg(
-        self, argname: str = None, argdesc: str = None, argtype: str = None, optarg: str = None, optshort: str = None
+        self,
+        argname: str | None = None,
+        argdesc: str | None = None,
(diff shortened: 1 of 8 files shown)


onenezicTT (approved):

ttexalens/elf_loader.py
```diff
@@ -213,35 +215,38 @@ def load_elf_sections(self, elf_path: str | ParsedElfFile, loader_data: str | in
...
-                        address = self.remap_address(address, loader_data_address, loader_code_address)
-                        read_data = self.read_block(address, len(data))
-                        if read_data != data:
-                            util.ERROR(f"Error writing section {name} to address 0x{address:08x}.")
-                            continue
-                        else:
-                            util.VERBOSE(
-                                f"Section {name} loaded successfully to address 0x{address:08x}. Size: {len(data)} bytes"
-                            )
+            if verify_write:
```
onenezicTT: does it make sense to do this check instead below line 215 after each write?

PR #848, opened 2026-01-15
Exposing remote devices connected to local device in Device class

### LINKED ISSUE #845: Make local device aware what remote devices it communicates with
We should add `cached_property` to device class for local devices which would return list of remote devices that are connected to that local device. This would make it easier to know what remote devices we should skip in `tt-triage` when local device breaks.

adjordjevic-TT:
Closes #845 

Added `cached_property` `remote_devices` in `Device` class that gets all remote devices that are connected to that local device. This can be used in `tt-triage` to easily know what remote devices we should skip when local device breaks.

---

> [!NOTE]
> Introduces a lightweight way to query remote devices tied to a local device.
> 
> - Adds cached property `Device.remote_devices` in `ttexalens/device.py` that returns `[device for device in context.devices if not device.is_local and self.id == cluster_descriptor.get_closest_mmio_capable_chip(device.id)]`
> - Enforces usage on local devices via assertion
> - Enables forward references with `from __future__ import annotations` to support `list[Device]` typing
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6d8656b12c56349d9e537010056ada264e9488e6. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


ttexalens/device.py
```diff
@@ -152,6 +153,17 @@ def firmware_version(self):
...
 
+    # Get all remote devices that are connected to this local device
+    @cached_property
+    def remote_devices(self) -> list[Device]:
+        assert self.is_local, "Only local devices can get remote devices"
+        return [
+            device
+            for device in self._context.devices.values()
+            if not device.is_local
+            and self.id == self._context.cluster_descriptor.get_closest_mmio_capable_chip(device.id)
```
tt-vjovanovic: You can use `local_device` property

```suggestion
            if device.local_device == self
```

I think I didn't want to specify `@cached_property` as UMD might change logic, but since we are not going to use many hoops to get to devices (like old Galaxy), I think you can also set `@cached_property` to `local_device` as well...

PR #856, opened 2026-01-16
Add safe mode to library noc reads and writes

onenezicTT:
Adds an optional "safe" flag in the library read and write methods which will enforce all the addresses from which we can read/write to be in known and allowed memory regions.

---

> [!NOTE]
> **Adds safe NoC read validation (default-on)**
> 
> - Introduces `validate_noc_read_is_safe` to verify that `addr..addr+size` lies within known, NoC-accessible memory blocks; raises `TTException` otherwise
> - Extends `read_from_device`, `read_word_from_device`, and `read_words_from_device` with `safe_mode=True` (optional) and invokes the validation before reads
> 
> **Tests**
> - New parameterized unit tests in `test_lib.py` ensuring `safe_mode=True` raises for invalid addresses across `read_from_device`, `read_word_from_device`, and `read_words_from_device`
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b137157184188233d3e0a502e93ca6ab27468ea2. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/test/ttexalens/unit_tests/test_lib.py
+++ b/test/ttexalens/unit_tests/test_lib.py
@@ -757,7 +757,7 @@ class TestSafeAccess(unittest.TestCase):
             "eth",
         ]
 
-        for _, device in self.context.devices.items():
+        for device_id, device in self.context.devices.items():
             # Collect test locations from all block types
             test_locations = []
             for block_type in all_block_types:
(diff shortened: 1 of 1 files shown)


ttexalens/tt_exalens_lib.py
```diff
@@ -104,13 +104,51 @@ def convert_coordinate(
...
+    """
+    noc_block = location.noc_block
+    noc_memory_map = noc_block.noc_memory_map
+
+    bytes_checked = 0
+    while bytes_checked < num_bytes:
+        curr_addr = addr + bytes_checked
+        memory_block_info = noc_memory_map.find_by_noc_address(curr_addr)
+        if not memory_block_info:
+            raise TTException(f"Address 0x{curr_addr:08X} is not in a known memory block for location {location}")
```
tt-vjovanovic: Should this be some `UnsafeAccessException`?

adjordjevic-TT (approved):

ttexalens/tt_exalens_lib.py
```diff
@@ -270,6 +371,9 @@ def write_to_device(
     if len(data) == 0:
         raise TTException("Data to write must not be empty.")
 
+    if safe_mode:
+        validate_noc_access_is_safe(coordinate, addr, len(data), True)
```
tt-vjovanovic:
```suggestion
        validate_noc_access_is_safe(coordinate, addr, len(data), is_write=True)
```

test/ttexalens/unit_tests/test_lib.py
```diff
@@ -726,6 +728,541 @@ def test_unaligned_write_private_memory(self, loc_str: str, risc_name: str):
...
+            # Should not raise exception
+            lib.write_to_device(location, address, data, safe_mode=True, context=self.context)
+        else:
+            # Verify block is readable
+            self.assertTrue(
+                block_info.is_safe_to_read(address, num_bytes), f"Block '{block_name}' should be readable for this test"
+            )
+            # Should not raise exception
+            result = lib.read_from_device(location, address, num_bytes=num_bytes, safe_mode=True, context=self.context)
+            self.assertEqual(len(result), num_bytes)
```
tt-vjovanovic: To be honest, this seems too hard to follow :D
I would expect much simpler tests and device targeted :)
For example, on wormhole, functional_worker, you can read 1,2,3 and you cannot read 4,5,6.
There is no need to repeat logic for verifying is memory access ok or not.

On the other hand, I would also add tests when there is transition across two blocks, again device targeted. When user specifies to read from block 1, but because of read bytes it ends up in block 2 (which can be readable or not)...

PR #861, opened 2026-01-19
Run tests with NOC1 in CI as well

tt-vjovanovic:
Adding ability to run tests using NOC1.
Making them run in CI as well.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -144,11 +144,15 @@ jobs:
           python3 -m unittest discover -v -t . -s test/app -p *test*.py
 
       - name: Run Python tests for ttexalens library (using NOC1)
+        # Disabling NOC1 tests on n300 board due to known bug causing host to reboot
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -139,10 +139,24 @@ jobs:
...
           path: ${{ steps.strings.outputs.test_report_path }}
 
-      - name: Run Python tests for TTExaLens app
+      - name: Run Python tests for TTExaLens app (using NOC0)
         run: |
           python3 -m unittest discover -v -t . -s test/app -p *test*.py
 
+      - name: Run Python tests for ttexalens library (using NOC1)
+        # Disabling NOC1 tests on n300 board due to known bug causing host to reboot
+        if: ${{ matrix.runner-info.board != 'n300' }}
```
adjordjevic-TT: If we now exactly which tests cause this maybe we can just skip those if NOC1 is used and board type is n300. Maybe not right now in this PR, but something to keep in mind when we have all info.

PR #911, opened 2026-02-24
Fix go noc command when failover is turned on

onenezicTT:
- When `go -n X` is ran from cli it will not be respected when failover is turned on so we always stay on that same noc, even though we falsely show in the cli that we switched NoCs.
- When a timeout happens during `device` command, for each cell we'll spam the device again and break it and kill the host, added support so that it stops at that point
- Noc failover can now be disabled in cli

### DIFF
--- a/ttexalens/cli.py
+++ b/ttexalens/cli.py
@@ -427,7 +427,9 @@ def main():
         server_ip = address[0] if address[0] != "" else "localhost"
         server_port = address[-1]
         util.INFO(f"Connecting to TTExaLens server at {server_ip}:{server_port}")
-        context = init_ttexalens_remote(server_ip, int(server_port), use_4B_mode, safe_mode=safe_mode, noc_failover=noc_failover)
+        context = init_ttexalens_remote(
+            server_ip, int(server_port), use_4B_mode, safe_mode=safe_mode, noc_failover=noc_failover
+        )
     else:
         context = init_ttexalens(
             init_jtag=init_jtag,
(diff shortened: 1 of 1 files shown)


ttexalens/cli_commands/device_summary.py
```diff
@@ -154,7 +155,9 @@ def cell_render_function(loc: OnChipCoordinate) -> str:
                         x = data & 0x3F
                         y = (data >> 6) & 0x3F
                         cell_contents_str.append(color_block(f"{x:02}-{y:02}", block_type))
-                    except:
+                    except TimeoutDeviceRegisterError:
+                        raise
+                    except Exception:
```
tt-vjovanovic: Maybe add debug/trace logging for exception text, so that we can easier debug...

ttexalens/device.py
```diff
@@ -524,9 +535,16 @@ def append_horizontal_axis_labels(rows, ui_hor_range):
                     if cell_renderer == None:
                         render_str = all_block_locs[(ui_hor, ui_ver)].to_str("logical")
                     else:
-                        render_str = cell_renderer(all_block_locs[(ui_hor, ui_ver)])
+                        try:
+                            render_str = cell_renderer(all_block_locs[(ui_hor, ui_ver)])
+                        except TimeoutDeviceRegisterError:
+                            render_aborted = True
+                            break
                 row.append(render_str)
```
tt-vjovanovic: (re)raise doesn't help?

onenezicTT: I was thinking about not raising to leave the cli usable, but since we broke the noc it doesn't make too much sense

PR #936, opened 2026-03-12
Disable BH DMA

tt-vjovanovic:
UMD wants to disable BH DMA because of hardware bug. Syseng is investigating mitigation. Currently, UMD is silently skipping DMA, but they want to change it to exception.

### DIFF
--- a/ttexalens/umd_device.py
+++ b/ttexalens/umd_device.py
@@ -86,6 +86,10 @@ class UmdDevice:
     def is_simulation(self) -> bool:
         return self._is_simulation
 
+    @property
+    def can_use_dma(self) -> bool:
+        return self._arch != tt_umd.ARCH.BLACKHOLE and self._is_mmio_capable and not self._is_simulation
+
     def __select_noc_id(self, noc_id: int):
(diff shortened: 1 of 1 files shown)


ttexalens/umd_device.py
```diff
@@ -137,7 +142,12 @@ def __read_from_device_reg(self, coord_x: int, coord_y: int, address: int, size:
 
     def __write_to_device_reg(self, coord_x: int, coord_y: int, address: int, data: bytes, dma_threshold: int):
         # Check if we can use DMA write
-        if len(data) >= dma_threshold and self._is_mmio_capable and not self._is_simulation:
+        if (
+            len(data) >= dma_threshold
+            and self._arch != tt_umd.ARCH.BLACKHOLE
+            and self._is_mmio_capable
+            and not self._is_simulation
```
miacim: Maybe to move and 
self._arch != tt_umd.ARCH.BLACKHOLE
            and self._is_mmio_capable
            and not self._is_simulation
to function can_use_dma() or similar

PR #950, opened 2026-03-20
Select simulation device type based on path extension

pjanevskiTT:
## Summary
- Select `TTSimTTDevice` for `.so` simulation paths (functional simulator), `RtlSimulationTTDevice` for directory paths (RTL simulator)
- Add `TTDevice` type annotation to fix mypy incompatible assignment error
- Bump tt-umd to 0.9.5.dev260402

## Test plan
- [x] Verify with RTL simulator directory paths
- [x] Verify with `.so` functional simulator paths
- [x] Run mypy type checking passes cleanly

### DIFF
--- a/ttexalens/requirements.txt
+++ b/ttexalens/requirements.txt
@@ -1,4 +1,4 @@
-tt-umd==0.9.4.260326
+tt-umd==0.9.5.dev260402
 deprecation>=2.1.0
 docopt>=0.6.2
 fastnumbers>=5.1.0
--- a/ttexalens/umd_api.py
+++ b/ttexalens/umd_api.py
@@ -80,7 +80,10 @@ class UmdApi:
 
         UmdApi.select_noc_id(1 if initialize_with_noc1 else 0)
         if simulation_directory is not None:
-            tt_device = tt_umd.create_simulation_tt_device(simulation_directory)
+            if simulation_directory.endswith(".so"):
+                tt_device = tt_umd.TTSimTTDevice(simulation_directory)
+            else:
+                tt_device = tt_umd.RtlSimTTDevice(simulation_directory)
             soc_descriptor = tt_device.get_soc_descriptor()
             # Fix for simulator
             for core in soc_descriptor.get_cores(tt_umd.CoreType.TENSIX):
--- a/ttexalens/umd_api.py
+++ b/ttexalens/umd_api.py
@@ -80,7 +80,7 @@ class UmdApi:
 
         UmdApi.select_noc_id(1 if initialize_with_noc1 else 0)
         if simulation_directory is not None:
-            tt_device = tt_umd.RtlSimulationTTDevice.create(simulation_directory)
+            tt_device = tt_umd.create_simulation_tt_device(simulation_directory)
             soc_descriptor = tt_device.get_soc_descriptor()
             # Fix for simulator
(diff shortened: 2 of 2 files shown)


onenezicTT (dismissed):

onenezicTT:
Do you need to bump umd for this?

fvranicTT:
> Do you need to bump umd for this?

I believe so, yes.

ttexalens/umd_api.py
```diff
@@ -80,7 +80,10 @@ def __init__(
 
         UmdApi.select_noc_id(1 if initialize_with_noc1 else 0)
         if simulation_directory is not None:
-            tt_device = tt_umd.RtlSimulationTTDevice.create(simulation_directory)
+            if simulation_directory.endswith(".so"):
```
pjanevskiTT, line 84: needs more changes, won't merge it yet

onenezicTT:
I'm not really up for us picking up some "dev" tagged umd version, especially as when we want to bump exalens in metal we will run into trouble. I'd be fine if we have a "dev" exalens branch that we can just use "dev" umd versions, but in official, probably not. @tt-vjovanovic thoughts?

pjanevskiTT:
@aleksamarkovicTT on advising on how to name UMD version

ttexalens/requirements.txt
```diff
@@ -1,4 +1,4 @@
-tt-umd==0.9.4.260326
+tt-umd==0.9.5.dev260402
```
fvranicTT, line 1: This is OK for testing, but I wouldn't want to ingest this into our testing harness unless it's an official package that follows semantic versioning.

ttexalens/umd_api.py
```diff
@@ -80,7 +80,11 @@ def __init__(
 
         UmdApi.select_noc_id(1 if initialize_with_noc1 else 0)
         if simulation_directory is not None:
-            tt_device = tt_umd.RtlSimulationTTDevice.create(simulation_directory)
+            tt_device: tt_umd.TTDevice
+            if simulation_directory.endswith(".so"):
```
pjanevskiTT, line 84: Had to make fix this way until the API of simulation TTDevice classes is fully the same as RTL and TTSIm TTDevice API

ttexalens/umd_api.py
pjanevskiTT: fixed with new UMD release

pjanevskiTT:
@fvranicTT confirmed TTSim works now, will merge this PR

tt-vjovanovic (approved):

ttexalens/umd_api.py
```diff
@@ -80,7 +80,11 @@ def __init__(
...
+                tt_device = tt_umd.TTSimTTDevice.create(simulation_directory)
+            else:
+                tt_device = tt_umd.RtlSimulationTTDevice.create(simulation_directory)
```
Coaz00, line 87: I prefer it this way, both are fine thou.
```suggestion
            tt_device: tt_umd.TTDevice = tt_umd.TTSimTTDevice.create(simulation_directory) if simulation_directory.endswith(".so") else tt_umd.RtlSimulationTTDevice.create(simulation_directory)
```

Coaz00 (approved):

ttexalens/umd_api.py
pjanevskiTT: I will remove this logic anyway, so will keep it as it is

PR #956, opened 2026-03-25
brxy and wxy optimisation

### LINKED ISSUE #534: Consolidate arguments for CLI commands
Some arguments require noc location (`brxy`, `wxy`, `dump_regfile`, etc.) and some are not (`rv`).
`go` command already saves device id and core locations in `UIState`.
Instead of requiring noc location we can use one that is available in `UIState`.

djankovicTT:
closes #534 Added the option of excluding core-loc argument from `brxy` and `wxy` commands while keeping the user experience the same, with original form and order of the arguments. Also added the usage of `for_each` function for getting the device/s

Update: also added
    `dr`: option of excluding the location and using the default one
    `noc`: the default noc is noc status
    `go`: noc-loc can now be specified without typing -l before it


ttexalens/cli_commands/write_xy.py
```diff
@@ -44,24 +49,33 @@ def print_a_write(core_loc_str: str, addr: int, val: int):
...
+            addr += 4
+
+    for device in dopt.for_each(CommonCommandOptions.Device, context, ui_state):
+        if args["-d"]:
+            util.INFO(f"Writing to device {device.id}")
+        if raw_core_loc:
+            core_loc = OnChipCoordinate.create(str(raw_core_loc), device=device)
+            core_loc_str = str(raw_core_loc)
+        else:
+            core_loc = ui_state.current_location.change_device(device)
```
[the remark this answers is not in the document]
djankovicTT: This line is there in case multiple devices were specified but the core location is the default (from uistate). It doesn't alter uistate, but rather makes a new OnChipCoordinate with the current location (from uistate) but on the forwarded device, if the forwarded device is a different one from the one from uistate. If the device is the same as in uistate no new objects are made, it just returns that existing one

adjordjevic-TT (approved):
Great work! Before you merge you need to update documentation. You can do that by running `make docs` locally with your changes and it will automatically update documentation file and you can just push it.

tt-vjovanovic (approved):

adjordjevic-TT (approved):

ttexalens/cli_commands/burst_read_xy.py
```diff
@@ -47,52 +49,97 @@
...
+            raise util.TTException("brxy: address omitted; give at least one positional argument (the address).")
+        # One positional argument -> addr only
+        case 1:
+            addr_str = slots[0]
+        # Two positional arguments -> noc-loc, addr or addr, word-count
+        case 2:
+            try:
+                noc_loc, addr_str = OnChipCoordinate.create(slots[0], device=device), slots[1]
+            except util.TTException:
+                addr_str, word_count_str = slots[0], slots[1]
```
tt-vjovanovic: For me putting `=` for single assignment (when possible) is more readable, but it is up to you

tt-vjovanovic:
```suggestion
                noc_loc = OnChipCoordinate.create(slots[0], device=device)
                addr_str = slots[1]
            except util.TTException:
                addr_str = slots[0]
                word_count_str = slots[1]
```

ttexalens/cli_commands/burst_read_xy.py
```diff
@@ -47,52 +49,97 @@
...
+    return noc_loc, addr, word_count
+
+
 def run(cmd_text: str, context: Context, ui_state: UIState):
     dopt = tt_docopt(command_metadata, cmd_text)
     args = dopt.args
 
-    location_str = args["<core-loc>"]
+    slots = _brxy_docopt_ordered_slots(args)
+    raw_noc_loc, addr_arg, word_count = _resolve_brxy_positionals(slots, ui_state.current_location.device)
```
tt-vjovanovic: In ultimate future, there could be devices that are not of the same time attached to the same machine. User could use `-d` and then specified location wouldn't be current compared to current device. Parsing it once is only performance improvement, but I think it is safer to do it in the device loop.

ttexalens/cli_commands/go.py
```diff
@@ -4,10 +4,14 @@
 """
 Usage:
     go [-m <mode>] [-n <noc>] [ -d <device> ] [ -l <loc> ]
```
tt-vjovanovic:
```suggestion
    go [-m <mode>] [-n <noc>] [ -d <device> ] [ -l <noc-loc> ]
```

ttexalens/cli_commands/noc.py
```diff
@@ -3,6 +3,7 @@
 
 """
 Usage:
+    noc [-d <device>] [--noc <noc-id>] [-l <loc>] [-s]
```
tt-vjovanovic, line 6: what does command with no verb do?

tt-vjovanovic (approved):

ttexalens/cli_commands/noc.py
djankovicTT: `noc` is a shortcut for `noc status`. Thanks for the heads up, I now added it to the description

ttexalens/cli_commands/burst_read_xy.py
djankovicTT: That's a good point. I now changed to checking the `noc-loc` with the first device from `dopt.for_each()` instead. I figured there is no need to check with every one of them while just resolving the arguments, since that would implicitly be done each time i try to create a new `OnChipLocation` later in executing the burst read.

Let me know if this solution is acceptable

miacim (approved):

PR #969, opened 2026-03-30
Adding support for template parameters

tt-vjovanovic:
Extending `CallstackEntry` with template parameters that are extracted from DIE.

Adding callstack tests for verifying template parameters extraction.

Fixing bug in `ElfVariable` that was doing reading/writing from/to bytes to utilize if type is signed.

Fixing bug in callstack.


test/ttexalens/unit_tests/test_lib.py
```diff
@@ -1637,11 +1646,59 @@ def test_top_callstack_optimized(self, recursion_count: int, expected_f1_on_call
             pc = self.risc_debug.read_gpr(32)
         callstack: list[CallstackEntry] = lib.top_callstack(pc, parsed_elf, None, self.context)
 
-        self.assertEqual(len(callstack), expected_f1_on_callstack_count + 2)
+        self.assertEqual(len(callstack), expected_f1_on_callstack_count + 1)
```
adjordjevic-TT: Why was this changed? Was this a bug?

PR #998, opened 2026-04-23
Fixing flaky CI due to umd bug

### LINKED ISSUE #999: New KMD version causes our tests to fail on blackhole
After bumping UMD our CI started to be flaky. NOC1 tests fail with `TimeoutDeviceRegisterError`. UMD team suspects this behavior is due to KMD putting device in low power mode when there are no running processes. When we access device, sometimes process of switching out of low power mode is not over and be hit above mentioned error. We mitigated this by starting `tt-exalens` before noc1 tests which takes out device from low power mode a little bit earlier and avoids error.

Will add blocked by UMD issue when one is created.

### LINKED ISSUE #2449: UMD bump breaking tt-exalens CI
Recent bump of UMD version (`0.9.4.260326` -> `0.9.5.dev260402`) with this [PR](https://github.com/tenstorrent/tt-exalens/commit/b5dbb3e5fa243916ba6c4c6bc983df1897fd8698) caused tt-exalens CI to fail.

Reading address `0xffb121b0`through NOC1 on blackhole p150b triggers our heuristic for bad device state (read takes unusually long time (more than 2ms) and returns `0xFFFFFFFF`). CI was run multiple times and failed every time. Unsure how this even got merged since this is required workflow.
Failed run: https://github.com/tenstorrent/tt-exalens/actions/runs/24194229234/job/70636524185
After reverting UMD bump run passed: https://github.com/tenstorrent/tt-exalens/actions/runs/24195806157/job/70625667836

Could not reproduce this locally.

adjordjevic-TT:
This closes https://github.com/tenstorrent/tt-umd/issues/2449

Our NOC1 tests became flaky on blackhole due to issue with new KMD that puts device in low power mode when no processes are running. The problem is when we try to start UMD it does not wait for device to leave low power mode and due to that we can start testing before device is ready hitting an error.

We mitigate this by starting exalens before running tests so device has enough time to leave low power mode. This is a mitigation and will be reverted when UMD implements a fix.

Issue for tracking real fix: #999

Reintroducing reverted commit: #950

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -146,13 +146,13 @@ jobs:
       - name: Run Python tests for ttexalens library (using NOC1)
         run: |
           export TTEXALENS_TESTS_USE_NOC1=1
-          python3 tt-exalens.py & # Workaround due to bug: #999
+          python3 tt-exalens.py < /dev/null & # Workaround due to bug: #999
           python3 -m unittest discover -v -t . -s test/ttexalens -p *test*.py
 
       - name: Run Python tests for TTExaLens app (using NOC1)
         run: |
           export TTEXALENS_TESTS_USE_NOC1=1
-          python3 tt-exalens.py & # Workaround due to bug: #999
+          python3 tt-exalens.py < /dev/null & # Workaround due to bug: #999
           python3 -m unittest discover -v -t . -s test/app -p *test*.py
 
       - name: Install wheel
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -146,11 +146,13 @@ jobs:
       - name: Run Python tests for ttexalens library (using NOC1)
         run: |
           export TTEXALENS_TESTS_USE_NOC1=1
+          python3 tt-exalens.py & # Workaround due to bug: #999
```
tt-vjovanovic: Are we sure that this is correct mitigation and that it will survive warm reset that we do in tests?

adjordjevic-TT: Yeah I was thinking about it, what Nenad told me is that warm reset will not kill this process but just make its memory mapping unusable, but since we only need process to live this should be enough.

tt-vjovanovic (approved):

PR #1005, opened 2026-05-04
Add performance counter debug registers

miacim:
Added missing debug registers used for accessing performance counters.


onenezicTT (approved):

onenezicTT:
@copilot can you cross reference these added registers with what's in ISA documentation or metal?

adjordjevic-TT (approved):
Do we have any way of testing this?

PR #1006, opened 2026-05-05
Adding rocket core support

adjordjevic-TT:
This PR starts support for rocket cores debugging.
Added `QuasarFunctionalOverlayBlock` class that contains 8 rocket cores and is part of `QuasarFunctionalWorkerBlock` alongside neo blocks and is referenced by `neo_id=None`.

Added list of registers that was found in metal [here](https://github.com/tenstorrent/tt-metal/tree/main/tt_metal/hw/inc/internal/tt-2xx/quasar/overlay/meta/registers).

Added `QuasarRocketCoreDebug` that derives from `RiscDebug` and currently supports:
- asserting/deasserting reset
- halting/continuing
- getting pc


ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -0,0 +1,104 @@
...
+
+# SMN_RISC_RESET_REG bit 17: debug module out-of-reset (active-low, 0 = in reset)
+DM_OUT_OF_RESET_BIT = 1 << 17
+
+# Abstract command (COMMAND) — Access Register, 64-bit, transfer=1, write=0, regno=dpc (0x7B1)
+ABSTRACTS_BUSY = 1 << 12
+CMD_READ_DPC = (3 << 20) | (1 << 17) | 0x7B1
+
+
+class QuasarRocketCoreDebug(BabyRiscDebug):
```
tt-vjovanovic: nit: We might want to have `RocketCoreDebug` that is not part of Quasar (similar to `BabyRiscDebug`) and have classes that are specific to architecture (like `QuasarRocketCoreDebug`).

tt-vjovanovic: `RocketCoreDebug` shouldn't inherit `BabyRiscDebug`, but `RiscDebug`. If there is some code that should be shared between these classes, it should be moved from `BabyRiscDebug` to `RiscDebug`...

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -0,0 +1,104 @@
...
+                self.register_store.write_register("TT_DEBUG_MODULE_APB_DMCONTROL", DMACTIVE | hartsel)
+                self.register_store.write_register("TT_DEBUG_MODULE_APB_COMMAND", CMD_READ_DPC)
+                while self.register_store.read_register("TT_DEBUG_MODULE_APB_ABSTRACTCS") & ABSTRACTS_BUSY:
+                    pass
+                lo = self.register_store.read_register("TT_DEBUG_MODULE_APB_DATA0")
+                hi = self.register_store.read_register("TT_DEBUG_MODULE_APB_DATA1")
+                return (hi << 32) | lo
+
+    def set_code_start_address(self, address: int | None) -> None:
+        self.baby_risc_info.set_code_start_address(self.register_store, address)
```
tt-vjovanovic: You need to handle `address is None` case...

ttexalens/hardware/quasar/rocket_core_debug.py
adjordjevic-TT: There is no shared code, derived from BabyRiscDebug by mistake.

adjordjevic-TT: Sure that makes sense, I would assume rocket cores will appear in future archs.

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -0,0 +1,106 @@
...
+    def ensure_debug_module_out_of_reset(self) -> Generator[None, Any, None]:
+        value = self.register_store.read_register("SMN_RISC_RESET_REG")
+        dm_was_in_reset = not bool(value & DM_OUT_OF_RESET_BIT)
+        if dm_was_in_reset:
+            self.register_store.write_register("SMN_RISC_RESET_REG", value | DM_OUT_OF_RESET_BIT)
+        try:
+            yield
+        finally:
+            if dm_was_in_reset:
+                self.register_store.write_register("SMN_RISC_RESET_REG", value)
```
tt-vjovanovic, line 53: This might be unsafe. If risc is in reset and you want to take it out of reset, you need to know exactly what it will execute. What we used to do before was to put risc in endless spin loop (`for(;;);`).

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -0,0 +1,106 @@
...
+    def halt(self) -> None:
+        with self.ensure_debug_module_out_of_reset():
+            if self.is_halted():
+                util.WARN(f"Halt: {self.risc_location.risc_name} at {self.risc_location.location} is already halted")
+                return
+            hartsel = self.baby_risc_info.risc_id << 16
+            self.register_store.write_register("TT_DEBUG_MODULE_APB_DMCONTROL", DMACTIVE | hartsel | HALTREQ)
+            self.register_store.write_register("TT_DEBUG_MODULE_APB_DMCONTROL", DMACTIVE | hartsel)
+            if not self.is_halted():
+                raise RiscHaltError(self.risc_location.risc_name, self.risc_location.location)
```
tt-vjovanovic, line 69: I don't think using `self.ensure_debug_module_out_of_reset()` makes sense. Imagine risc is in reset, we will take it out of reset, halt it and put it back to reset.

tt-vjovanovic: Same goes with other methods.

tt-vjovanovic (changes requested):

ttexalens/hardware/quasar/rocket_core_debug.py
adjordjevic-TT: This is not putting risc in and out of reset, we just enable/disable debug module. Certain bit needs to be set in order to use debug module, this function enusres that debug module is out of reset when we try to use it.

ttexalens/hardware/quasar/rocket_core_debug.py
adjordjevic-TT: Can I just add if risc in reset skip all of this?

ttexalens/hardware/quasar/rocket_core_debug.py
tt-vjovanovic: Ah, my bad. Then we probably need to figure out what that exactly means. Will `ebreak` work if debug module is not enabled? What about if I want to debug something and put break/watch point? Does it require debug module to be enabled? We probably need to test a bit to understand behavior. For now, leave as is and we can improve later with more info.

ttexalens/hardware/quasar/rocket_core_debug.py
tt-vjovanovic: It was because of my misunderstanding for `ensure_debug_module_out_of_reset` => I was under impression it is `ensure_risc_out_of_reset`. Keep as is until we understand limitations.

tt-vjovanovic (approved):

PR #1021, opened 2026-05-13
hardware/baby_risc_info: fix JAL offset calculation in set_code_start_address

Dev-X25874:
## Summary

`set_code_start_address` falls back to writing a JAL instruction at
`default_code_start_address` when the core cannot change its start address
via a register. JAL in RISC-V is PC-relative, so the immediate must be
`target - PC`. The code was passing the raw `address` as the immediate,
which is only accidentally correct when `default_code_start_address == 0`
(BRISC). For all other cores the generated JAL would jump to
`default_code_start_address + address` instead of `address`.

## Root Cause

```python
# Before (incorrect for any non-zero default_code_start_address)
jump_instruction = ElfLoader.get_jump_to_offset_instruction(address)

# After (correct PC-relative offset)
jump_instruction = ElfLoader.get_jump_to_offset_instruction(address - self.default_code_start_address)
```

## Affected Cores

| Core   | `default_code_start_address` | Broken before fix? |
|--------|-----------------------------|--------------------|
| BRISC  | `0x00000000`                | No (offset == address) |
| TRISC0 | `0x00006000`                | ✅ Yes |
| TRISC1 | `0x0000A000`                | ✅ Yes |
| TRISC2 | `0x0000E000`                | ✅ Yes |
| NCRISC | `0x00012000`                | ✅ Yes |
| ERISC  | `0xFFC00000`                | ✅ Yes |

## Testing

Existing unit tests pass. The fix is a single-token change with no
API surface or behaviour change for BRISC (`0 - 0 == 0`).

### DIFF
--- a/ttexalens/hardware/baby_risc_info.py
+++ b/ttexalens/hardware/baby_risc_info.py
@@ -68,7 +68,7 @@ class BabyRiscInfo(RiscInfo):
 
             # If we cannot change the code start address, we write a jump instruction to the specified address
             assert self.default_code_start_address is not None
-            jump_instruction = ElfLoader.get_jump_to_offset_instruction(address)
+            jump_instruction = ElfLoader.get_jump_to_offset_instruction(address - self.default_code_start_address)
             register_store.location.noc_write32(self.default_code_start_address, jump_instruction)
         elif address is not None:
             if self.code_start_address_enable_register is not None and self.code_start_address_enable_bit is not None:


ttexalens/hardware/baby_risc_info.py
```diff
@@ -68,7 +68,7 @@ def set_code_start_address(self, register_store: RegisterStore, address: int | N
 
             # If we cannot change the code start address, we write a jump instruction to the specified address
             assert self.default_code_start_address is not None
-            jump_instruction = ElfLoader.get_jump_to_offset_instruction(address)
```
tt-vjovanovic, line 71: BRISC should be the only RISC that cannot change its default_code_start_address (it is hardcoded to 0, that's why this code works). Can you explain which risc doesn't work for you? Can you add `assert self.default_code_start_address == 0` and see what is happening?

Dev-X25874:
Good point — you're right that today only BRISC reaches this path, 
so the bug is latent rather than actively breaking anything.

That said, the assert self.default_code_start_address == 0 approach 
concerns me — it would silently make this path permanently 
BRISC-only at the code level, and the current implementation gives 
no indication that the fallback is fundamentally broken for any 
non-zero start address.

The fix costs one arithmetic operation and makes the code 
correct-by-construction: if a future core or config ever hits this 
path with a non-zero default, it will just work rather than silently 
jumping to the wrong address. The assert would turn that into a hard 
crash with no fallback.

Happy to add a comment documenting the BRISC-only assumption if 
that makes the intent clearer, but I'd argue the subtraction is 
still the right thing to do.

tt-vjovanovic:
> Good point — you're right that today only BRISC reaches this path, so the bug is latent rather than actively breaking anything.
> 
> That said, the assert self.default_code_start_address == 0 approach concerns me — it would silently make this path permanently BRISC-only at the code level, and the current implementation gives no indication that the fallback is fundamentally broken for any non-zero start address.
> 
> The fix costs one arithmetic operation and makes the code correct-by-construction: if a future core or config ever hits this path with a non-zero default, it will just work rather than silently jumping to the wrong address. The assert would turn that into a hard crash with no fallback.
> 
> Happy to add a comment documenting the BRISC-only assumption if that makes the intent clearer, but I'd argue the subtraction is still the right thing to do.

I thought you had a problem with our loader so I proposed `assert` in order to understand which risc is problematic. Fix is correct, just wasn't needed with current architectures.

tt-vjovanovic (approved):

Dev-X25874:
Thanks for the review and for merging, @tt-vjovanovic! Appreciated the discussion around the assert approach — it helped sharpen the reasoning behind the fix.

PR #1025, opened 2026-05-15
added search_memory() in tt_exalens_lib.py

djankovicTT:
- Added `search_memory()` function in `tt_exalens_lib.py` for searching for patterns in noc/risc private memory
- Updated `search-memory` CLI command to use it with no change to its behaviour or user experience
- Added tests for the new function and its helper functions in `test_lib.py`


ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +292,120 @@ def write_to_device(
...
+            break
+        matches.append(base_addr + idx)
+        if max_results is not None and len(matches) >= max_results:
+            break
+        start = idx + 1
+    return matches
+
+
+@trace_api
+def search_noc_memory(
```
tt-vjovanovic: Expose function in `ttexalens/__init__.py`

ttexalens/cli_commands/search-memory.py
```diff
@@ -39,15 +39,14 @@
...
-import math
-
 from ttexalens.context import Context
 from ttexalens.device import Device
 from ttexalens.uistate import UIState
 from ttexalens.coordinate import OnChipCoordinate
 from ttexalens import util as util
 from ttexalens.command_parser import CommandMetadata, tt_docopt, CommonCommandOptions
 from ttexalens.cli_commands.read import execute_safe_read, execute_unsafe_read
+from ttexalens.tt_exalens_lib import _auto_pattern_width, _build_pattern_bytes, _find_in_data
```
tt-vjovanovic: You should use method that you added to library and not private methods.

ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +292,120 @@ def write_to_device(
...
+    location: str | OnChipCoordinate,
+    pattern_values: list[int],
```
tt-vjovanovic: `pattern_values` should be something that user would set when calling this function in library.
Example:
```
search_memory("0,0", 0x1234)   # Search for single word value
search_memory("0,0", [0x1234, 0xacafaca])  # Search for list of words
search_memory("0,0", b"112233") # Search for explicit bytes in memory
```
Also, please rename to `pattern`.

ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +292,120 @@ def write_to_device(
...
+    start_addr: int = 0,
+    end_addr: int | None = None,
+    width: int | str = "auto",
+    max_results: int | None = 10,
+    device_id: int = 0,
+    context: Context | None = None,
+    chunk_size: int = 0x100000,
+) -> list[tuple[int, str]]:
```
tt-vjovanovic: Ideally, not all results should be returned. Imagine that user wanted to search through DRAM block (2GB) and pattern was too common and it should return 1000000 entries.
Consider changing this API to return address of first result and then user can call again with `start_address` after that result to continue search. If users want all results (for better perf), we can always add new function that would return all results, but usually first results is good enough...

ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +292,120 @@ def write_to_device(
...
+    return matches
+
+
+@trace_api
+def search_noc_memory(
+    location: str | OnChipCoordinate,
+    pattern_values: list[int],
+    start_addr: int = 0,
+    end_addr: int | None = None,
+    width: int | str = "auto",
```
tt-vjovanovic: I don't think here `width` is necessary. Look at other methods in this file. They choose whether list element is one byte or word...

adjordjevic-TT (commented):
Add tests to `test_lib.py`

ttexalens/tt_exalens_lib.py
djankovicTT: I initially removed the option of passing `None` to the `max_results` argument (which would return all found search results) but I see that the CLI `search` command has a `--max-results all` option which would want all. 
Do you then suggest I add a new separate function to the lib that i would call from `search` if `all` is passed to the `--max-results` argument, or revert to again accepting `None` in the lib function and it meaning to return all results, but leave the default of the argument at `1` (result) for this intended behaviour?

djankovicTT: Currently `--max-results all` option doesn't work in CLI since CLI calls lib functions and the option of all isn't supported there now
So just let me know if you think it should: 

- Stay that way (so I remove option `all` from the Description of the command), or
- Add an additional lib function that would return all search results, or
- Bring back the `None` option for max-results argument in lib function and make that mean to return all (like before) but keep the default at `1`

ttexalens/cli_commands/search-memory.py
```diff
@@ -10,14 +10,13 @@
...
+  --end=<end>        End address (exclusive) or 'all'. If omitted, searches all contiguous blocks from --start.
                      'all' searches all blocks from --start onwards.
   --width=<width>    Bytes per pattern element (any positive integer) or auto. [default: auto]
                      auto = minimum power-of-2 byte count needed to represent the largest element,
                      applied uniformly to all elements.
-  --read-size=<rs>   Maximum bytes per device read. Defaults to 4 when -r is used (RISC debug hardware),
-                     or 1MB otherwise.
-  --max-results=<n>  Maximum number of matches to return, or 'all'. [default: 1]
+  --read-size=<rs>   Maximum bytes per device read. Defaults to 1MB.
+  --max-results=<n>  Maximum number of matches to return. [default: 1]
```
adjordjevic-TT: Consider making default a little bigger, I would assume user would want all of matches displayed if there is reasonable amount of them. I would probably put soemthing like 10.

ttexalens/cli_commands/search-memory.py
```diff
@@ -284,22 +143,21 @@ def run(cmd_text: str, context: Context, ui_state: UIState):
...
-    else:
-        try:
-            max_results = int(max_results_arg)
-        except ValueError:
-            util.ERROR(f"Invalid --max-results value: {max_results_arg!r}. Must be a positive integer or 'all'.")
-            return
-        if max_results < 1:
-            util.ERROR(f"--max-results must be at least 1, got {max_results_arg!r}.")
-            return
+    try:
```
adjordjevic-TT: You can do this a little bit cleaner. In try after converting `max_results_arg` to integer check if it is less than 1 and if so throw `ValueError`. This way we only print error in one spot avoid code duplication and text of first error message is good enough for both cases.

ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +291,200 @@ def write_to_device(
...
+    return matches
+
+
+@trace_api
+def search_noc_memory(
+    location: str | OnChipCoordinate,
+    pattern: int | list[int] | bytes,
+    start_addr: int = 0,
+    end_addr: int | None = None,
+    max_results: int = 1,
```
adjordjevic-TT: Same comment for max_results argument, consider making it bigger.

ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +291,200 @@ def write_to_device(
...
+            prev_tail = search_data[-overlap:] if overlap > 0 else b""
+            chunk_addr += len(chunk_bytes)
+
+        current = block_end
+
+    return all_matches
+
+
+@trace_api
+def search_riscv_memory(
```
adjordjevic-TT: There is a lot of code duplication in `search_noc_memory` and `search_riscv_memory` maybe make helper function to avoid that.

ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +291,203 @@ def write_to_device(
...
+    noc_id = check_noc_id(None, coordinate.context)
+    use_4B_mode = check_4B_mode(None, coordinate.context)
+    memory_map = coordinate.noc_block.noc_memory_map
+    all_matches: list[tuple[int, str]] = []
+    current = start_addr
+
+    while end_addr is None or current < end_addr:
+        block_info = memory_map.find_by_noc_address(current)
+        if block_info is None:
+            break
```
tt-vjovanovic: We should do the same check before the `while` loop and probably throw exception so that user is aware that it is targeting wrong address. Imagine scenario that user made a mistake by specifying search range as `0xFFAFFFF0` instead of `0xFFB00000` and it returns empty result instead of telling user that start address is wrong and nothing has been searched.

ttexalens/tt_exalens_lib.py
```diff
@@ -291,6 +291,203 @@ def write_to_device(
...
+        ):
+            return all_matches
+
+        current = block_end
+
+    return all_matches
+
+
+@trace_api
+def search_riscv_memory(
```
tt-vjovanovic: Do we need two methods? If we add `risc_name` and `neo_id` as optional arguments to previous method, can it do work for both ones? What these methods high level do:
- figure out which block it should read
- read chunk of data from the block
- search for pattern in that data (taking care about switching from chunk to chunk)
So basically all we need to solve is how do we read data and which memory_map we are using.

ttexalens/tt_exalens_lib.py
tt-vjovanovic: I think I would go with following solution:
- remove `max_results` argument
- always return all results from last read chunk
Then user will control results count with `chunk_size`.
This way, we are performant if we need to return all results (all chunks are read only once). If CLI user wants to see `all` results we will just repeatedly call this method until we process memory user requested.

ttexalens/cli_commands/search-memory.py
```diff
@@ -10,14 +10,13 @@
...
+  --end=<end>        End address (exclusive) or 'all'. If omitted, searches all contiguous blocks from --start.
                      'all' searches all blocks from --start onwards.
   --width=<width>    Bytes per pattern element (any positive integer) or auto. [default: auto]
                      auto = minimum power-of-2 byte count needed to represent the largest element,
                      applied uniformly to all elements.
-  --read-size=<rs>   Maximum bytes per device read. Defaults to 4 when -r is used (RISC debug hardware),
-                     or 1MB otherwise.
-  --max-results=<n>  Maximum number of matches to return, or 'all'. [default: 1]
+  --read-size=<rs>   Maximum bytes per device read. Defaults to 1MB.
+  --max-results=<n>  Maximum number of matches to return. [default: 10]
```
tt-vjovanovic: We should revert to previous behavior.

ttexalens/cli_commands/search-memory.py
```diff
@@ -10,14 +10,13 @@
 
 Options:
   --start=<start>    Start address for the search. Defaults to 0, or to the start of RISC private memory if -r is specified.
-  --end=<end>        End address (exclusive) or 'all'. If omitted, searches only the block containing --start.
+  --end=<end>        End address (exclusive) or 'all'. If omitted, searches all contiguous blocks from --start.
```
tt-vjovanovic: Why would we want to search all contiguous blocks? They hold very different data, most of them cannot be read with more than 4 bytes as they contain registers.

ttexalens/tt_exalens_lib.py
djankovicTT: You're right that is an important edge case to consider, thank you for pointing that out!
I will add this check and a test for it as well

djankovicTT:
Changes to `search_memory()` lib function:

- no `max_results` argument - instead it returns **all** results from the searched memory range
-  `end_addr`- the one to dictate where the search stops: 
    - `None` (default) - search only the block containing start_addr.
    - int ≥ 0               - exclusive end address; may span multiple blocks if the range crosses a block boundary.
    - "all"                  - search all contiguous blocks from start_addr.
- `chunk_size`      - still the size of one read, defaults to 4 bytes for risc private memory and 1MB for noc

I was hesitant to make `chunk_size` the limiting size of the whole search, since I thought the user might not know that that is the case, and when let's say he calls the function and gets "no results found" he might not know that just a small chunk of memory was actually searched. Also, isn't `end_address` technically not needed if you expect to always search `chunk_size` memory or even if you search more than one `chunk_size` due to the `end_address`, you suggested the function returns just results from the last chunk (why are we then searching anything other than the last chunk)?

These are my thoughts:
- ether leave it like I did now - the lib function searches ether from `start_addr` to `end_addr` if `end_addr` is forwarded, or from `start_addr` to the end of that block if `end_addr` is omitted; all in increments of `chunk_size`. Then the CLI just limits how much it wants to show based on `--max-results` argument it has, or
- I rename `chunk_size` to something like `search_size` and possibly make it a mandatory argument so the user knows how much memory he's searching, and if so - also possibly remove the `end_addr` argument so the responsibility of arguments is clearer to the user

@tt-vjovanovic please review and let me know you're thoughts :)

tt-vjovanovic:
> I was hesitant to make `chunk_size` the limiting size of the whole search, since I thought the user might not know that that is the case, and when let's say he calls the function and gets "no results found" he might not know that just a small chunk of memory was actually searched. Also, isn't `end_address` technically not needed if you expect to always search `chunk_size` memory or even if you search more than one `chunk_size` due to the `end_address`, you suggested the function returns just results from the last chunk (why are we then searching anything other than the last chunk)?

Returning results from read last chunk doesn't mean it is last chunk of data that will be read. It is first chunk that contains result and you will return data from it. You can also return "next search start address" for performance (if last chunk contains single result that is first 8 bytes, and you don't want to read again 1MB that doesn't contain search pattern).

Example:
```
[c0] [c1] [c2] [c3] [c4] [c5] [c6] ... [cn]
```
`[c3]` has result and `[cn-1]` also contains result. First search will return all results from chunk `[c3]` and tell that "next search start address" should be start of `[c4]`. Then if user calls search again with address of `[c4]` it will return all results from `[cn-1]`.

ttexalens/cli_commands/search-memory.py
```diff
@@ -29,13 +29,13 @@
...
   search 0xDEADBEEF                              # Search for pattern in the block containing address 0
-  search 0xDEADBEEF --end all                    # Search all accessible blocks
-  search 0xDEADBEEF --start 0x10000 --end all    # Search all blocks from 0x10000 onwards
+  search 0xDEADBEEF --end all                    # Search all contiguous blocks from address 0
+  search 0xDEADBEEF --start 0x10000 --end all    # Search all contiguous blocks from 0x10000 onwards
   search 0x1234 --start 0 --end 0xFFFF           # Search from address 0 to 0xFFFF
   search 0x1234 0x5678 --width 2                 # Search for two 2-byte LE values (0x34 0x12 0x78 0x56)
   search 0xDEADBEEF --read-size 64               # Search using 64-byte reads
   search 0xAB --unsafe                           # Search with safety checks bypassed
-  search 0xBEEF -r brisc                         # Search brisc private memory (4-byte reads by default)
```
tt-vjovanovic, line 38: This example comment was useful for users to understand what happens when they use risc name in command

tt-vjovanovic: Also, return code that was setting `--read-size`.

ttexalens/tt_exalens_lib.py
```diff
@@ -292,6 +292,221 @@ def write_to_device(
...
+        return next_chunk_addr if next_chunk_addr < end_addr else None
+    # end_addr == "all": check whether a contiguous block starts at next_chunk_addr.
+    if risc_name is not None:
+        memory_map = coordinate.noc_block.get_risc_debug(risc_name, neo_id).risc_info.memory_map
+        return next_chunk_addr if memory_map.find_by_private_address(next_chunk_addr) is not None else None
+    return (
+        next_chunk_addr
+        if coordinate.noc_block.noc_memory_map.find_by_noc_address(next_chunk_addr) is not None
+        else None
+    )
```
tt-vjovanovic: I was expecting this to be single expression (`last_tested_address+1`) or at max one if testing if it is last chunk (`last_tested_address+1` escapes valid block or `end_addr`). How it turned into ~30 lines of code?

ttexalens/tt_exalens_lib.py
```diff
@@ -292,6 +292,221 @@ def write_to_device(
...
+    start = 0
+    while True:
+        idx = data.find(pattern, start)
+        if idx == -1:
+            break
+        matches.append(base_addr + idx)
+        if max_results is not None and len(matches) >= max_results:
+            break
+        start = idx + 1
+    return matches
```
tt-vjovanovic: Please delete this method. It creates additional array that you "convert" immediately into new one. `max_results` is not being used. I don't see this method being useful.

djankovicTT: Thanks for the heads up - it's leftover from when i had max_results as a lib function argument. Now that it's not, this method has no value at all - will delete it!

ttexalens/tt_exalens_lib.py
```diff
@@ -292,6 +292,184 @@ def write_to_device(
     coordinate.noc_write(addr, data, noc_id, use_4B_mode, safe_mode=safe_mode)
 
 
+def _encode_pattern(pattern: int | list[int] | bytes) -> bytes:
```
onenezicTT, line 295: These helper methods do not look like they should belong in this file

ttexalens/tt_exalens_lib.py
```diff
@@ -292,6 +292,184 @@ def write_to_device(
...
+            range_end = block_end if effective_end is None else min(block_end, effective_end)
+            read_fn = lambda addr, size: coordinate.noc_read(addr, size, noc_id, use_4B_mode, safe_mode=safe_mode)
+            yield current, range_end, block_info.name, read_fn
+            if end_addr is None:
+                return
+            current = block_end
+
+
+@trace_api
+def search_memory(
```
onenezicTT, line 406: What's the value in adding this to our API? Do we have scenarios that require explicit library call vs calling the cli command as it's also callable by code?

test/ttexalens/unit_tests/test_lib.py
```diff
@@ -1727,3 +1728,188 @@ def test_callstack_invalid(self, location, elf_paths, offsets=None, risc_name="b
         # Check for invalid location
         with self.assertRaises((TTException, ValueError, FileNotFoundError)):
             lib.callstack(location, elf_paths, offsets, risc_name, None, max_depth, True, device_id, self.context)
+
+
+class TestSearchHelpers(unittest.TestCase):
```
onenezicTT, line 1733: Do these tests bring any value, I would think not really?

PR #1037, opened 2026-05-23
Make ELF parsing thread-safe

tt-vjovanovic:
When using tt-triage with multithreaded access the biggest problem is ParsedElf that is not thread safe. The reason for this is using elftools classes. Solution: protect access to any elftools class with reentrant lock that is added in ParsedElf class.


ttexalens/elf/parsed.py
```diff
@@ -131,77 +132,92 @@ def decode_file_line(dwarf: DWARFInfo) -> dict[int, tuple[str, int, int]]:
...
+        with self._parsed_elf._lock:
+            return self._section.data()
 
 
 class ParsedElfFile:
     def __init__(self, elf: ELFFile, elf_file_path: str):
         self.elf = elf
         self.elf_file_path = elf_file_path
         self.loaded_offset = 0
+        self._lock = threading.RLock()
```
onenezicTT, line 65: Why `RLock` vs `lock`?

onenezicTT (approved):

ttexalens/elf/parsed.py
tt-vjovanovic: Because of iterators. We take lock there and then in some other function that would do data caching...

PR #1041, opened 2026-05-25
Fixing inconsistency in taking rocket core's debug module out of reset

adjordjevic-TT:
To ensure debug module is ready for use after taking it out of reset we need to set `DMCONTROL` to 1 and wait for `TT_CLUSTER_CTRL` register to be set or we might try using debug module before it is ready.

### DIFF
--- a/ttexalens/hardware/quasar/rocket_core_debug.py
+++ b/ttexalens/hardware/quasar/rocket_core_debug.py
@@ -71,6 +71,8 @@ class QuasarRocketCoreDebug(RocketCoreDebug):
             hartsel = self.baby_risc_info.risc_id << 16
             self.register_store.write_register("TT_DEBUG_MODULE_APB_DMCONTROL", DMACTIVE | hartsel | HALTREQ)
             self.register_store.write_register("TT_DEBUG_MODULE_APB_DMCONTROL", DMACTIVE | hartsel)
+            if not self.is_halted():
+                raise RiscHaltError(self.risc_location.risc_name, self.risc_location.location)
 
     def cont(self) -> None:
         with self.ensure_debug_module_out_of_reset():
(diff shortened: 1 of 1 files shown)


ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -46,6 +46,12 @@ def ensure_debug_module_out_of_reset(self) -> Generator[None, Any, None]:
         dm_was_in_reset = not bool(value & DM_OUT_OF_RESET_BIT)
         if dm_was_in_reset:
             self.register_store.write_register("SMN_RISC_RESET_REG", value | DM_OUT_OF_RESET_BIT)
+            self.register_store.write_register("TT_DEBUG_MODULE_APB_DMCONTROL", 1)
```
miacim, line 49: I would rather have this in more descriptive function like enable_debug_module

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -46,6 +46,12 @@ def ensure_debug_module_out_of_reset(self) -> Generator[None, Any, None]:
...
+            while self.register_store.read_register("TT_CLUSTER_CTRL_DEBUG_DMACTIVE") != 1:
+                pass
```
miacim, line 51: I expect that this is waiting for debug module to become active. Shouldn't we put sleep() and some additional exception logic. It also might be ok to be in separate function.
Probably self.register_store.write_register("TT_CLUSTER_CTRL_DEBUG_DMACTIVEACK", 1) should be part of that function also.

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -46,6 +46,12 @@ def ensure_debug_module_out_of_reset(self) -> Generator[None, Any, None]:
...
+            self.register_store.write_register("TT_CLUSTER_CTRL_DEBUG_DMACTIVEACK", 1)
+            value = self.register_store.read_register("SMN_RISC_RESET_REG")
```
miacim, line 53: This is repeated on many places. Maybe consider having function is_debug_module_out_of_reset()

miacim (approved):

Issue #1058, opened 2026-06-03
Reading/writing via hart not available for Rocket cores


adjordjevic-TT:
Reading from and writing to memory via hart is an optional feature of overlay block's debug module which is used to perform reads and writes exactly like Rocket core would which guarantees us data coherency. This feature is not implemented for Quasar. We need to be able to read fresh data from Rocket core's private memory in order to get full callstacks and variables while program is hung.

Now we are implementing workaround using debug module to perform reads and writes via system bus. It is still unclear if this guarantees data coherency and issue is open for that investigation #1057.

If workaround turns out to provide coherent reads we can lower this issue priority, if opposite is true we need to raise this and see if we can get it for Quasar or at least for next generation.

adjordjevic-TT:
Reading and writing via system bus is confirmed (as explained in this issue #1057) to be coherent so reading or writing via hart is not needed.

Issue #1057, opened 2026-06-03
Investigate whether reads/writes via system bus are coherent


adjordjevic-TT:
We are implementing reading from and writing to Rocket core's private memory using debug module to perform reads and writes via system bus. We need to investigate whether these are coherent or in other words are we guaranteed to get fresh data.

We will confirm this by creating waveform of these reads and writes and from that waveform hardware team should be able to tell us if there are coherent or not.

adjordjevic-TT:
Waveform sent, hardware team is adding test to verify data coherency for this approach.

adjordjevic-TT:
Waiting for this to be merged to close this issue https://yyz-gitlab.local.tenstorrent.com/tensix/soc/overlay/-/merge_requests/708

PR #1060, opened 2026-06-03
Enabling reading/writing via debug modul's system bus

### LINKED ISSUE #980: Read/write Rocket core's private memory
Each Rocket core has instruction and data cache (L1-I and L1-D, each 4kB). Also, all rocket cores share 128kB L2 cache and also share TL1 memory with Tensixes. If we try to read from TL1 we risk having stale data (data from L1/L2 was not yet flushed to TL1). If we want to issue transactions through NoC/SMN we need to explicitly flush ensure fresh data. Other approach is to access private memory of Rocket cores via hart (Rocket core itself is performing load/store instructions giving us the same view as if we were running code on it). To achieve that we need to do the following:
1. Halt (or ebreak)
2. Write address into `DATA` at `0x0300A010`
3. Issue a command at `0x0300A05C`
4. Wait until busy at `0x0300A05C` is 0
5. Value is now in `DATA`

adjordjevic-TT:
Closes #980 

Enabling reading/writing via debug modul's system bus in overlay block. This should allow us coherent access to rocket core's private memory (this is still under investigation #1057). If this does not happen to provide coherent access we need to investigate possibility of adding reading via hart #1058.

### DIFF
--- a/ttexalens/hardware/quasar/rocket_core_debug.py
+++ b/ttexalens/hardware/quasar/rocket_core_debug.py
@@ -7,7 +7,6 @@ from typing import Any, Generator
 import time
 
 from ttexalens import util
-from ttexalens.device import Device
 from ttexalens.exceptions import RiscHaltError
 from ttexalens.hardware.baby_risc_info import BabyRiscInfo
 from ttexalens.hardware.rocket_core_debug import RocketCoreDebug
@@ -37,10 +36,6 @@ class QuasarRocketCoreDebug(RocketCoreDebug):
     def __init__(self, risc_info: BabyRiscInfo, register_store: RegisterStore, enable_asserts: bool = True):
         super().__init__(risc_info, enable_asserts)
         self.register_store = register_store
-
-    @property
-    def device(self) -> Device:
-        return self.risc_info.noc_block.device
 
     def is_in_reset(self) -> bool:
         reset_bit = 1 << self.baby_risc_info.reset_flag_shift
--- a/ttexalens/hardware/quasar/rocket_core_debug.py
+++ b/ttexalens/hardware/quasar/rocket_core_debug.py
@@ -24,6 +24,13 @@ DM_OUT_OF_RESET_BIT = 1 << 17
 ABSTRACTS_BUSY = 1 << 12
 CMD_READ_DPC = (3 << 20) | (1 << 17) | 0x7B1
 
+# System Bus Access Control and Status (SBCS) fields.
+SBCS_SBBUSY = 1 << 21
+SBCS_SBREADONADDR = 1 << 20
+SBCS_SBACCESS_32 = 2 << 17
+SBCS_SBERROR_MASK = 0x7 << 12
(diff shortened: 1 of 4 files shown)


ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -120,5 +127,35 @@ def get_pc(self) -> int:
...
+
+    def read_memory(self, address: int, safe_mode: bool | None = None) -> int:
+        """Read a 32-bit word via System Bus Access."""
+        with self.ensure_debug_module_out_of_reset():
+            self._sba_wait_not_busy()
+            self.register_store.write_register(
+                "TT_DEBUG_MODULE_APB_SBCS",
+                SBCS_SBACCESS_32 | SBCS_SBREADONADDR | SBCS_SBERROR_MASK | SBCS_SBBUSYERROR,
+            )
+            self.register_store.write_register("TT_DEBUG_MODULE_APB_SBADDR0", address & 0xFFFFFFFF)
```
onenezicTT: Consider whether it makes sense to have some sort of assert inside `read/write_register` on 32 bit values or some util like ` _int32(val)`. Just thinking out loud

onenezicTT (approved):

ttexalens/hardware/quasar/rocket_core_debug.py
adjordjevic-TT: Yes this probably makes more sense than cutting address to 32 bits.

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -120,5 +127,35 @@ def get_pc(self) -> int:
...
+        """Poll SBCS until the system bus manager is idle. Raise on timeout."""
+        start_time = time.time()
+        while self.register_store.read_register("TT_DEBUG_MODULE_APB_SBCS") & SBCS_SBBUSY:
+            if time.time() - start_time > timeout:
+                raise Exception("Timeout waiting for system bus access")
+            time.sleep(0.01)
+
+    def read_memory(self, address: int, safe_mode: bool | None = None) -> int:
+        """Read a 32-bit word via System Bus Access."""
+        with self.ensure_debug_module_out_of_reset():
```
tt-vjovanovic: I keep seeing this method `ensure_debug_module_out_of_reset()` and thinking that it should be something different :D
Can we rename it to `ensure_debug_module_is_active()` or something along those lines?

ttexalens/hardware/quasar/rocket_core_debug.py
tt-vjovanovic: Same on other places. Consider having utility function that will do this check and throw something meaningful. We might want to use it on other places as well (like previous RiscDebug methods), or even have something like `_read_memory()`/`_write_memory()` that won't do these checks, while original `read_memory()`/`write_memory()` would do the check and call new "protected" methods that we can override in Wormhole/Blackhole/Quasar classes.

tt-vjovanovic (approved):

PR #1063, opened 2026-06-04
Remove GDB binary from wheel packaging

tt-vjovanovic:
Eliminate the GDB binary from the Python wheel packaging process. Update the logic for retrieving the GDB client to ensure it is located at runtime rather than bundled, enhancing portability across platforms. Adjust related documentation to reflect these changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,14 +37,12 @@ tt-exalens = "ttexalens.cli:main"
 
 [tool.scikit-build]
 cmake.build-type = "Release"
-cmake.args = ["-DBUILD_PYTHON_WHEEL=ON", "-DSTRIP_SYMBOLS=ON"]
+cmake.args = ["-DBUILD_PYTHON_WHEEL=ON"]
 wheel.install-dir = "ttexalens"
 wheel.packages = ["ttexalens"]
 wheel.py-api = "py3"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -37,14 +37,12 @@ tt-exalens = "ttexalens.cli:main"
...
 cmake.build-type = "Release"
-cmake.args = ["-DBUILD_PYTHON_WHEEL=ON", "-DSTRIP_SYMBOLS=ON"]
+cmake.args = ["-DBUILD_PYTHON_WHEEL=ON"]
 wheel.install-dir = "ttexalens"
 wheel.packages = ["ttexalens"]
 wheel.py-api = "py3"
 wheel.platlib = false
 experimental = true
-# Install the python_wheel component for wheel packaging
-install.components = ["python_wheel"]
```
onenezicTT, line 47: Why remove this?

onenezicTT (approved):

adjordjevic-TT (approved):

pyproject.toml
tt-vjovanovic: That was CMake step that was copying gdb client on wheel path. Look at `CMakeLists.txt:45`.

PR #1068, opened 2026-06-05
Refactor coverage integration and improve gcov handling

### LINKED ISSUE #567: Use libgcov directly
We link with code from `riscv-src/coverage/gcov.c` when building kernels for coverage. That file contains code from GCC's tree: `gcc/libgcc/libgcov-driver.c`, but stripped of everything we don't need (we only need `__gcov_info_to_gcda` and its dependencies).

If we link with `libgcov.a` normally, too many things from libgcov, and even newlib, gets pulled into the final binary.

An investigation is needed. Excerpt from `nm` output for `libgcov.a` (sfpi 6.11.1, based on GCC 12.4.0):
`
_gcov_info_to_gcda.o:
00000070 T __gcov_info_to_gcda
         U __udivsi3
         U __umodsi3
         U abort
00000028 t dump_counter
00000000 t dump_unsigned
00000000 b list_size_length.0
00000004 b list_sizes.1
`

It could be that libc gets pulled in due to `abort`. If that's the case, it needs to be seen what calls it (there's no direct reference to it in the source) and if anything can be done about it. Another possibility is that this will be solved if libgcov gets built with `-ffunction-sections` (and the related `-fdata-sections`, but that's less needed) which will get rid of unneeded transitive dependencies at link time (if we call ld with `--gc-sections`).

tt-vjovanovic:
Enhance the integration of coverage functionality by refactoring the handling of gcov. This update streamlines the coverage support file and optimizes the build process for coverage-related libraries. Additionally, it improves the handling of line information in the Dwarf debugging format.

Closes #567.


onenezicTT (approved):

riscv-src/coverage.c
```diff
@@ -5,10 +5,17 @@
...
 #include <stddef.h>
 #include <stdint.h>
 
-#include "gcov.h"
+// Needed functions and types for libgcov. These are the only ones that are actually called by libgcov.
+typedef int64_t gcov_type;
+void __gcov_merge_add(gcov_type* counters, unsigned n_counters) {}
+int gcov_error(const char* fmt, ...) { return 0; }
+void abort() { while (1); }
+uint32_t __umodsi3(uint32_t a, uint32_t b) { return a % b; }
```
iklikovacTT: These `a % b` and `a / b` compile into calls of `__umodsi3` and `__udivsi3`, so you get infinite recursion. Check the objdump:
```
000008c4 <__umodsi3>:
 8c4:   ff010113                addi    sp,sp,-16
 8c8:   00112623                sw      ra,12(sp)
 8cc:   ff9ff0ef                jal     8c4 <__umodsi3>
 8d0:   00c12083                lw      ra,12(sp)
 8d4:   01010113                addi    sp,sp,16
 8d8:   00008067                ret

000008dc <__udivsi3>:
 8dc:   ff010113                addi    sp,sp,-16
 8e0:   00112623                sw      ra,12(sp)
 8e4:   ff9ff0ef                jal     8dc <__udivsi3>
 8e8:   00c12083                lw      ra,12(sp)
 8ec:   01010113                addi    sp,sp,16
 8f0:   00008067                ret
```

iklikovacTT: I suggest just removing those definitions and letting GCC supply its own primitives. The `g++` driver includes `-lgcc` on the command line, so it should just work out of the box.

riscv-src/coverage.c
```diff
@@ -55,7 +62,7 @@ static void write_data(const void* _data, unsigned int length, void* arg) {
     *written += length;
 }
 
-static size_t strlen(const char* s) {
+size_t strlen(const char* s) {
```
iklikovacTT, line 67: GCC at -O3 sees that this loop computes strlen, and compiles it into a call to strlen, which is, well, this function itself. The resulting code:
```
000008f4 <strlen>:
 8f4:   00054783                lbu     a5,0(a0)
 8f8:   02078263                beqz    a5,91c <strlen+0x28>
 8fc:   ff010113                addi    sp,sp,-16
 900:   00150513                addi    a0,a0,1
 904:   00112623                sw      ra,12(sp)
 908:   fedff0ef                jal     8f4 <strlen>
 90c:   00c12083                lw      ra,12(sp)
 910:   00150513                addi    a0,a0,1
 914:   01010113                addi    sp,sp,16
 918:   00008067                ret
 91c:   00000513                li      a0,0
 920:   00008067                ret
```
is correct in the abstract sense, but recurses for every character in the string.
`-fno-builtin-strlen` or, more broadly, `-fno-builtin`, makes GCC emit the intended code.

iklikovacTT: Interestingly, GCC seems to have peeled off the first iteration of the loop in one pass (for n=0, which became the base case), then realized in a subsequent pass that the resulting loop looks similar to strlen (more accurately, it computes `strlen(s+1)`). It then emitted a call to the library strlen (this function itself) and adjusted the result. Looks to me like it emitted functioning code by, more or less, coincidence.

Regardless, the fix is pretty simple.

riscv-src/coverage.c
```diff
@@ -5,10 +5,17 @@
...
+#include <gcov.h>
 #include <stddef.h>
 #include <stdint.h>
 
-#include "gcov.h"
+// Needed functions and types for libgcov. These are the only ones that are actually called by libgcov.
+typedef int64_t gcov_type;
+void __gcov_merge_add(gcov_type* counters, unsigned n_counters) {}
+int gcov_error(const char* fmt, ...) { return 0; }
+void abort() { while (1); }
```
iklikovacTT: If I'm not mistaken, this file is built as C++17? If so, `while(1);` is undefined behavior.
```suggestion
void abort() { for(;;) asm volatile(""); }
```

ttexalens/cpp/elf/dwarf_info.cpp
```diff
@@ -31,76 +32,31 @@ DwarfInfo::DwarfInfo(DwarfInfo&&) noexcept = default;
...
-        Dwarf_Unsigned ln = 0;
-        Dwarf_Unsigned col = 0;
-        DwarfString src(dbg);
-        dwarf_lineno(match, &ln, &error);
-        dwarf_lineoff_b(match, &col, &error);
-        dwarf_linesrc(match, &src, &error);
-        return DwarfFileLine{std::string(src.get()), static_cast<uint32_t>(ln), static_cast<uint32_t>(col)};
+    const auto& ranges = impl->get_line_ranges();
+
+    // Ranges are sorted by `low` and disjoint (see LineRange), so the only
```
iklikovacTT: Can we enforce this invariant somehow?

Off the top of my head, inlining, LTO and COMDAT could cause overlapping ranges.

iklikovacTT (changes requested):

riscv-src/coverage.c
iklikovacTT: That wasn't introduced by this PR (in fact, that may well could've been my mistake when I originally worked on coverage, can't remember), but I believe it's worth fixing here while you're at it.

PR #1081, opened 2026-06-13
Move get_callstack and get_frame_callstack to C++

tt-vjovanovic:
Moving `get_callstack` and `get_frame_callstack` to C++ code so other tools can use it. Needed for LLK sanitizer.


ttexalens/tt_exalens_lib.py
```diff
@@ -624,28 +644,29 @@ def callstack(
...
     if risc_debug.is_in_reset():
         raise TTException(f"RiscV core {risc_debug.risc_location} is in reset")
-    return risc_debug.get_callstack(elfs, offsets, max_depth, stop_on_main)
+
+    with risc_debug.ensure_halted():
+        # Reading the program counter from risc register
+        pc = risc_debug.read_gpr(32)
+
+        # If ebreak was hit, pc will point to the instruction after it
+        if risc_debug.is_ebreak_hit():
```
adjordjevic-TT, line 661: This might not be true for rocket cores, but I guess we will cross that bridge when we come to it.

adjordjevic-TT (approved):

ttexalens/tt_exalens_lib.py
tt-vjovanovic: That is correct. We should probably move this check to `RiscDebug` specialization classes (in this case `BabyRiscDebug`).

Issue #528, opened 2025-07-30
Writing/reading trisc2 private memory through debug hardware fails periodically on blackhole


adjordjevic-TT:
**Problem:**
Writing/reading TRISC2 private memory through debug hardware sometimes fails on `blackhole`, specifically at addresses with a remainder of 8 or 12 when divided by 16. Here’s part of the output from a test that writes to TRISC2 private memory and then reads it back.
```
0xffb00000: PASSED
0xffb00004: PASSED
0xffb00008: FAILED -> 0x0 != 0x12345678
0xffb0000c: FAILED -> 0x0 != 0x12345678
0xffb00010: PASSED
0xffb00014: PASSED
0xffb00018: FAILED -> 0x0 != 0x12345678
0xffb0001c: FAILED -> 0x0 != 0x12345678
```
This pattern continues through whole private memory.
**TODO:**
Investigate why this is happening.
Create waveform and get more info from hardware team.
Repeat same test on quasar.

**Completion criteria:**
Either bug fix or skipping failing addresses with proper explanation.

adjordjevic-TT:
The same test was run on the Quasar simulator, where it executed successfully and all tested addresses passed.

PR #1089, opened 2026-06-17
Implement tail call frame reconstruction and inline frame expansion

tt-vjovanovic:
Introduce functionality for reconstructing tail call frames and expanding inline frames during callstack retrieval. This enhancement allows for a more accurate representation of the source-level call chain, improving debugging capabilities. Additionally, the option to expand tail call inline frames is now available in the API.


test/ttexalens/unit_tests/test_lib.py
```diff
@@ -1634,6 +1634,116 @@ def test_callstack_namespace(self, elf_name):
...
+    @parameterized.expand(CALLSTACK_ELFS)
+    def test_callstack_tail_call(self, elf_name):
+        # 0xFFFFFFFA selects the tail_call_test chain in callstack.cc. In the optimized build ra1
+        # tail-calls pc1 (the call becomes a jump), so the ra1 frame - with its inlined ra2/ra3 -
+        # leaves no return address on the stack and is only recoverable from the DWARF call-site
+        # (tail-call) information. In the -O0 (debug / coverage) build the same logical stack is
+        # produced through ordinary unwinding. GDB reconstructs the tail-called frame in both cases,
+        # so it is the ground truth our walker must match regardless of optimization level.
+        if self.device.is_blackhole() and self.risc_name == "trisc2":
+            self.skipTest("This test doesn't work as expected due to blackhole trisc2 hardware bug, tt-exalens:#528")
```
adjordjevic-TT, line 1646: We are referencing this #528 issue in few places, but that issue is closed. I assume we closed it when we found workaround by reading private memory directly via NOC but since we disabled that maybe we should reopen this issue?

adjordjevic-TT (approved):

test/ttexalens/unit_tests/test_lib.py
tt-vjovanovic: That issue describes that the problem is. It will never get fixed, it is hardware issue. Having it open doesn't make sense. Not sure if we should update something in that issue (@miacim).

PR #1091, opened 2026-06-19
Various small perf improvements

tt-vjovanovic:
- Changed how Pyro5 native objects are exposed. Previously we would create new type for every object. Now we create type for every type and just instantiate for objects.
- Changed how we enumerate GDB server available processes in tests. If `debug_only_with_elfs` is specified, we don't iterate over riscs to look for elfs, but other way around: we iterate over elfs and then check riscs.
- Replacing how we choose NOC ID in `umd_device.py` and `umd_api.py`. It is faster to cache it in Python TLS than to go all the way to C++. Will be changed with next UMD bump (end of next week, or the week after)
- If 4 byte mode is disabled, there is no need to do timeout detection (as timings are wrong).
- Settings code start address in baby risc info is changing configuration register. It is faster to read value (as we can do this over debug registers) and then to update only if it has changed.
- Don't go over all sections in elf loader. Use `get_section_by_name()` instead.
- Caching `noc_block` property in `OnChipCoordinate`.
- Updating GDB memory access tests to use `skip_detach` and `debug_only_with_elfs` as they are not testing general GDB server behavior.
- Removing `device()` property in `OnChipCoordinate` as it is being called 900K times in tt-triage on T3K. It was only forwarding to field, so upgrading field to be public.

### DIFF
--- a/ttexalens/gdb/gdb_server.py
+++ b/ttexalens/gdb/gdb_server.py
@@ -21,7 +21,7 @@ from ttexalens.gdb.gdb_data import GdbProcess, GdbThreadId
 from ttexalens.gdb.gdb_file_server import GdbFileServer
 from ttexalens.context import Context
 from ttexalens import util as util
-from ttexalens.hardware.risc_debug import RiscLocation
+from ttexalens.hardware.risc_debug import RiscDebug, RiscLocation
 from ttexalens.memory_access import RestrictedMemoryAccessError
 
 # Helper class returns currently debugging list of threads to gdb client in paged manner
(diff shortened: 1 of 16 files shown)


ttexalens/umd_device.py
```diff
@@ -290,23 +303,25 @@ def __write_to_device_reg_unaligned_helper(
         dma_threshold: int,
     ):
         size_in_bytes = len(data)
+        do_write = self.__write_to_device_reg_no_timeout if use_4B_mode else self.__write_to_device_reg
```
adjordjevic-TT: Why are we using no timeout version if in 4B mode? Couldn't this crash host?

ttexalens/server.py
```diff
@@ -80,70 +81,87 @@ def stop(self):
...
+    def _wrap_result(self, result):
+        """Prepares a method/property result for transport over Pyro5.
+
+        Native types and known serializable types are returned as-is so Pyro5/serpent
+        forwards them directly. Any other object is wrapped, registered with the daemon
+        (once per object), and returned as a proxy.
+        """
+        result_type = type(result)
+        # Check if result_type is simple type
+        if result_type in (int, float, str, bool, type(None), list, dict, set, tuple, bytes):
```
adjordjevic-TT: Maybe it would be cleaner to make a constant `SIMPLE_TYPES` or something like that for this tuple.

ttexalens/gdb/gdb_server.py
```diff
@@ -110,37 +110,47 @@ def available_processes(self):
...
-                        )
-                        process = GdbProcess(pid, elf_path, risc_debug, virtual_core, block_type)
-                    else:
-                        process = last_process
-                    available_processes[process.process_id] = process
-                    last_available_processes[risc_debug.risc_location] = process
+
+        def add_process(risc_debug: RiscDebug, elf_path: str | None):
+            # Check if process is in self._last_available_processes and reuse it if it is
+            # TODO: In ideal world, we would have "start time" for a core (time when core was taken out of reset) and use that as a key for reusing process id; for now, we can just check if elf path is the same
```
adjordjevic-TT, line 116: Do you have any ideas on how we can know this? If another process is putting cores in and out of reset I do not see how we can keep track of that.

adjordjevic-TT (approved):

ttexalens/gdb/gdb_server.py
tt-vjovanovic: That is TODO from very long time. We should revisit it when Quasar hardware is here and we have debugging hardware that works (or we think it works).

ttexalens/umd_device.py
tt-vjovanovic: Good catch, one `not` is missing :D

PR #1097, opened 2026-06-26
Making NOC1 default

### LINKED ISSUE #850: Move default noc reads from noc0 to noc1
Breaking noc0 could break device reset and force us to reset the host so to be more safe by default we'll use noc1. We could potentially only move triage's initialization of exalens to noc1 instead of forcing all library users that currently maybe rely on exalens working in noc0 to switch to noc1

adjordjevic-TT:
Closes #850

Making NOC1 default noc. As described in the issue mentioned above this is safer since reset is using NOC0 so breaking it is much worse than breaking NOC1.

Updated documentation.

Now we use NocId from UMD instead of having use_noc1 true/false.


test/ttexalens/unit_tests/test_base.py
```diff
@@ -60,12 +59,12 @@ def init_cached_test_context():
     return _cached_test_context
 
 
-def init_test_context(use_noc1: bool = False, safe_mode: bool = False):
-    if use_noc1:
-        assert not os.getenv("TTEXALENS_TESTS_REMOTE"), "Remote testing for NOC1 not supported"
-        return init_ttexalens(use_noc1=True, noc_failover=False, safe_mode=safe_mode)
+def init_test_context(noc_id: NocId | None = None, safe_mode: bool = False):
+    if noc_id == NocId.NOC1 or noc_id == NocId.SMN:
```
tt-vjovanovic: What about when `nod_id == None`?

test/ttexalens/unit_tests/test_noc.py
```diff
@@ -5,24 +5,24 @@
...
 @parameterized_class(
     [
-        {"use_noc1": False, "noc_id": 0},
-        {"use_noc1": False, "noc_id": 1},
-        {"use_noc1": True, "noc_id": 0},
-        {"use_noc1": True, "noc_id": 1},
+        {"init_noc_id": NocId.NOC0, "noc_id": 0},
+        {"init_noc_id": NocId.NOC0, "noc_id": 1},
+        {"init_noc_id": NocId.NOC1, "noc_id": 0},
+        {"init_noc_id": NocId.NOC1, "noc_id": 1},
```
tt-vjovanovic: Are these supposed to run on SMN on Quasar?

ttexalens/cli_commands/go.py
```diff
@@ -12,7 +12,7 @@
     noc-loc     Optional. X-Y or R,C, or dram channel (e.g. ch3). Use interchangeably with -l <loc>.
 
 Options:
-    -n <noc>     Use NOC1 or NOC0 for communication with the device. [0: NOC0, 1: NOC1]
+    -n <noc>     NOC to use for communication with the device. [0: NOC0, 1: NOC1, 2: SMN]
```
tt-vjovanovic: I'm not sure if we want to add SMN for WH/BH. Similarly NOC1 doesn't exist on Quasar.
What about this command having "Default NOC" (NOC0) and "Alternative NOC" (NOC1 or SMN)?

ttexalens/hardware/blackhole/device.py
```diff
@@ -44,6 +44,7 @@ class BlackholeDevice(Device):
     def __init__(self, id: int, umd_device: UmdDevice, context: Context):
         super().__init__(id, umd_device, context)
         self.instructions = BlackholeInstructions()
+        self._noc_to_use = [context.noc_id, NocId.NOC0 if context.noc_id != NocId.NOC0 else NocId.NOC1]
```
tt-vjovanovic: I think `Device` object should define list of available NOCs and failover should go through the list. So, `Device` would define `self.available_nocs = [NocId.NOC0, NocId.NOC1]`, then during failover, we know which noc was chosen and we can go to the next one in the list.

ttexalens/hardware/quasar/device.py
```diff
@@ -29,6 +29,7 @@ class QuasarDevice(Device):
 
     def __init__(self, id: int, umd_device: UmdDevice, context: Context):
         super().__init__(id, umd_device, context)
+        self._noc_to_use = [context.noc_id, NocId.NOC0 if context.noc_id != NocId.NOC0 else NocId.SMN]
```
tt-vjovanovic: Same comment as in `BlackholeDevice`

ttexalens/hardware/quasar/noc_block.py
```diff
@@ -26,11 +27,11 @@ def __init__(
         self.register_store_noc = RegisterStore(default_niu_register_store_initialization_noc, self.location)
         self.register_store_smn = RegisterStore(default_niu_register_store_initialization_smn, self.location)
 
-    def get_register_store(self, noc_id: int = 0, neo_id: int | None = None) -> RegisterStore:
+    def get_register_store(self, noc_id: int = int(NocId.NOC0), neo_id: int | None = None) -> RegisterStore:
```
tt-vjovanovic: Why `get_register_store` method is not using `NocId`, but it is using `int`?

ttexalens/hardware/wormhole/device.py
```diff
@@ -40,6 +40,7 @@ class WormholeDevice(Device):
     def __init__(self, id: int, umd_device: UmdDevice, context: Context):
         super().__init__(id, umd_device, context)
         self.instructions = WormholeInstructions()
+        self._noc_to_use = [context.noc_id, NocId.NOC0 if context.noc_id != NocId.NOC0 else NocId.NOC1]
```
tt-vjovanovic: Same comment as in `BlackholeDevice` class.

ttexalens/hardware/perf_counters.py
```diff
@@ -172,11 +172,10 @@ def read_counter(
 
     def _register_store(self, noc_id: int | None) -> RegisterStore:
```
tt-vjovanovic: `noc_id` should be of `NocId` type.

ttexalens/_lib_helpers.py
```diff
@@ -55,11 +55,11 @@ def check_context(context: Context | None = None) -> Context:
     return init.GLOBAL_CONTEXT
 
 
-def check_noc_id(noc_id: int | None, context: Context) -> int:
+def check_noc_id(noc_id: NocId | int | None, context: Context) -> int:
```
tt-vjovanovic: I think this method should return `NocId`.

ttexalens/cli.py
```diff
@@ -242,7 +242,7 @@ def get_dynamic_prompt() -> HTML:
                             if ui_state.gdb_server.is_connected:
                                 gdb_status += "(connected)"
                             my_prompt += f"gdb:{gdb_status} "
-                        noc_prompt = f"{ui_state.current_device.active_noc}"
+                        noc_prompt = f"{int(ui_state.current_device.active_noc)}"
```
tt-vjovanovic: If `active_noc` is `NocId`, maybe we want not `int` value. Figure out what makes more sense.

ttexalens/cli.py
```diff
@@ -390,7 +390,7 @@ def main():
         util.VERBOSE(f"Verbosity level: {util.Verbosity.get().name} ({util.Verbosity.get().value})")
 
     safe_mode = False if args["--unsafe-mode"] else True
-    use_noc1 = args["--use-noc1"]
+    noc_id = NocId(int(args["--noc-id"]))
```
tt-vjovanovic: This might give bad parsing errors

ttexalens/context.py
```diff
@@ -21,14 +22,21 @@
     from ttexalens.server import FileAccessApi
     from ttexalens.umd_api import UmdApi
 
+
+class NocId(IntEnum):
```
tt-vjovanovic: Is there some enumeration type in UMD API?

ttexalens/context.py
```diff
@@ -41,24 +49,27 @@ def __init__(
         self.dma_write_threshold: int = dma_write_threshold
 
         self.noc_failover = noc_failover
-        self._use_noc1 = use_noc1
+        # Current noc selected for communication.
+        self._noc_id = noc_id
+        # Noc selected for initialization. Will be removed when #1102 is resolved.
+        self.init_noc_id = noc_id
```
tt-vjovanovic:
```suggestion
        self.init_noc_id = noc_id # TODO #1102: Noc selected for initialization.
```

ttexalens/context.py
```diff
@@ -41,24 +49,27 @@ def __init__(
         self.dma_write_threshold: int = dma_write_threshold
 
         self.noc_failover = noc_failover
-        self._use_noc1 = use_noc1
+        # Current noc selected for communication.
+        self._noc_id = noc_id
```
tt-vjovanovic:
```suggestion
        self._noc_id = noc_id
```

ttexalens/device.py
```diff
@@ -371,6 +371,15 @@ def arc_msg(
 
     def read_arc_telemetry_entry(self, noc_id: int | None, telemetry_tag: int) -> int:
         def noc_operation(noc_id: int) -> int:
+            # ARC telemetry must be read over the NOC selected at initialization (#1102), regardless of the
+            # context's current noc_id (which may have been changed after init).
```
tt-vjovanovic:
```suggestion
            # TODO #1102: ARC telemetry must be read over the NOC selected at initialization, regardless of the
            # context's current noc_id (which may have been changed after init).
```

ttexalens/tt_exalens_lib.py
```diff
@@ -354,8 +354,8 @@ def arc_msg(
         list[int]: return code, reply0, reply1.
     """
     context = check_context(context)
-    noc_id = check_noc_id(noc_id, context)
     device = validate_device_id(device_id, context)
+    noc_id = check_noc_id(noc_id, context)
```
tt-vjovanovic, line 358: Since `check_noc_id` now accepts `NocId` as well, we should add `NocId` as acceptable type for all methods in this file

ttexalens/umd_api.py
```diff
@@ -49,7 +50,7 @@ def create_simulation_cluster_descriptor(arch: tt_umd.ARCH) -> str:
 @Pyro5.api.expose
 class UmdApi:
     @staticmethod
-    def select_noc_id(noc_id: int, arch: tt_umd.ARCH | None = None):
+    def select_noc_id(noc_id: NocId | int, arch: tt_umd.ARCH | None = None):
```
tt-vjovanovic: Internal API mush use `NodId`. There should be conversion only for library methods.

ttexalens/umd_device.py
```diff
@@ -53,15 +54,16 @@ def __init__(
     @staticmethod
     def initialize_device_coords_cache(
         soc_descriptor: tt_umd.SocDescriptor, arch: tt_umd.ARCH
-    ) -> list[list[list[tt_umd.CoreCoord | None]]]:
+    ) -> dict[NocId, list[list[tt_umd.CoreCoord | None]]]:
```
tt-vjovanovic: Using dictionary makes this function so much slower than list, and it is being called millions of times. What this method should use is index in `Device` object array of `available_nocs`.
We will eventually remove this method, we are waiting for UMD to property implement noc switching

test/ttexalens/unit_tests/test_base.py
adjordjevic-TT: We always pass noc_id when using this function so I am removing option of having `None` value.

ttexalens/hardware/quasar/noc_block.py
adjordjevic-TT: I missed it, changing it.

ttexalens/context.py
adjordjevic-TT: There is and it is exposed via nanobind, I am switching to using that.

ttexalens/cli.py
adjordjevic-TT: Yeah, I switched it to print enum field name and removed `noc:` part so now we will get `NOC1` for example.

ttexalens/cli.py
adjordjevic-TT: What do you mean exactly?

ttexalens/cli_commands/go.py
adjordjevic-TT: Yeah this will break stuff if we try initializing context with SMN on blackhole/wormhole this will raise some UMD  exception. The problem is before initializing context we do not know device type therefore cannot say if SMN is valid or not for example. Default NOC can be same for everyone but alternate NOC is different for wh/bh and quasar which runs into same problem as described above.

I talked with UMD guys and they say on silicon it is possible to know the architecture before doing topology discovery so that might be the way. On simulation I am not sure how we can know this, maybe a hack can be reading it from `simulation_directory` since it contains arch name (or am I wrong here and this is not always the case)?

test/ttexalens/unit_tests/test_noc.py
adjordjevic-TT: I would avoid testing SMN on Quasar since it is super slow, I need to think about how to make these noc tests architecture aware.

test/ttexalens/unit_tests/test_base.py
```diff
@@ -60,12 +59,12 @@ def init_cached_test_context():
     return _cached_test_context
 
 
-def init_test_context(use_noc1: bool = False, safe_mode: bool = False):
-    if use_noc1:
-        assert not os.getenv("TTEXALENS_TESTS_REMOTE"), "Remote testing for NOC1 not supported"
-        return init_ttexalens(use_noc1=True, noc_failover=False, safe_mode=safe_mode)
+def init_test_context(noc_id: NocId, safe_mode: bool = False):
+    if noc_id == NocId.NOC1 or noc_id == NocId.SYSTEM_NOC:
```
tt-vjovanovic: Hm... This doesn't seem right. Only if you set `noc_id` to be `NocId | None` then you can test for its value; if it is not `None` then user explicitly set it and that is not supported.

ttexalens/cli_commands/go.py
```diff
@@ -40,10 +40,10 @@ def run(cmd_text: str, context: Context, ui_state: UIState):
 
     if args["-n"] is not None:
         noc = int(args["-n"])
-        if noc not in [0, 1]:
-            util.ERROR("NOC must be 0 or 1")
+        if noc not in [0, 1, 2]:
+            util.ERROR("NOC must be 0 (NOC0), 1 (NOC1), or 2 (SYSTEM_NOC)")
```
tt-vjovanovic: Consider allowing users to type `NOC0`, `SYSTEM_NOC`, `SMN`, etc.

ttexalens/cli_commands/tensix_reg.py
```diff
@@ -79,7 +79,7 @@ def run(cmd_text: str, context: Context, ui_state: UIState):
     value_str: str | None = None
     data_type: REGISTER_DATA_TYPE | None = None
     register_pattern: str | None = dopt.args["<register_pattern>"] if dopt.args["--search"] else None
-    noc_id: int = dopt.args["-n"] if dopt.args["-n"] else 0
+    noc_id: NocId = to_noc_id(int(dopt.args["-n"])) if dopt.args["-n"] else NocId.NOC0
```
tt-vjovanovic: Hm... Probably we need a method to be shared for parsing `NocId`...

ttexalens/hardware/blackhole/arc_block.py
```diff
@@ -106,10 +107,10 @@ def __init__(self, location: OnChipCoordinate):
             ],
         )
 
-    def get_register_store(self, noc_id: int = 0, neo_id: int | None = None) -> RegisterStore:
-        if noc_id == 0:
+    def get_register_store(self, noc_id: NocId = NocId.NOC1, neo_id: int | None = None) -> RegisterStore:
```
tt-vjovanovic: Why don't we default to `self.device.active_noc` instead of hard coding differently for other architectures?

tt-vjovanovic: Other places as well.

ttexalens/register_store.py
```diff
@@ -87,9 +87,11 @@ def bar0_address(self) -> int | None:
         return self.base_address.bar0_address + self.offset
 
     @property
-    def noc_id(self) -> int | None:
+    def noc_id(self) -> NocId | None:
         assert self.base_address is not None and self.base_address.noc_address is not None
-        return self.base_address.noc_id
+        # DeviceAddress.noc_id is the raw NOC address-space index (0/1); a register living in a given NOC's
+        # address space must be accessed over that same NOC, so map it to the corresponding NocId.
+        return to_noc_id(self.base_address.noc_id) if self.base_address.noc_id is not None else None
```
tt-vjovanovic: Update `DeviceAddress.noc_id` to be `NocId`.

tt-vjovanovic (approved):

ttexalens/hardware/blackhole/arc_block.py
adjordjevic-TT: Yeah this makes more sense, done!

test/ttexalens/unit_tests/test_noc.py
adjordjevic-TT: Will do it in another PR since right now we do not run these tests anyway on quasar

PR #1103, opened 2026-07-01
Avoid reading internal pipeline state in LLK math_reconfig.py test

mcraigheadTT:
This LLK test reads back debug daisychain state before and after to verify that it is undisturbed. Some of the (daisy_sel, signal_sel) values it is reading are real (persistent) state of the ADCs and RWCs; others are internal, transient, undocumented RTL pipeline signals that do not reflect real state, and checking their equivalence before/after is not meaningful. Limit the reads to only the persistent ADC and RWC (including fidelity_phase) state.

Aside from being a general code hygiene improvement for silicon, this change also fixes these ~2400 tests on ttsim, which cannot return meaningful values for these daisy_sel/signal_sel pairs.

### DIFF
--- a/ttexalens/hardware/blackhole/functional_worker_debug_bus_signals.py
+++ b/ttexalens/hardware/blackhole/functional_worker_debug_bus_signals.py
@@ -2125,6 +2125,6 @@ group_map: dict[str, tuple[int, int]] = {
 }
 
 tensix_debug_bus_description = TensixDebugBusDescription(
-    register_window_counter_groups=[group_name for group_name in group_map.keys() if group_name.startswith("rwc")],
-    address_counter_groups=[group_name for group_name in group_map.keys() if group_name.startswith("adc")],
+    register_window_counter_groups=["rwc_coordinates_a", "rwc_coordinates_b", "rwc_fidelity_phase"],
+    address_counter_groups=[group_name for group_name in group_map.keys() if group_name.startswith(("adcs0", "adcs2"))],
 )
--- a/ttexalens/hardware/wormhole/functional_worker_debug_bus_signals.py
+++ b/ttexalens/hardware/wormhole/functional_worker_debug_bus_signals.py
@@ -930,6 +930,6 @@ group_map = {
 }
 
 tensix_debug_bus_description = TensixDebugBusDescription(
-    register_window_counter_groups=[group_name for group_name in group_map if group_name.startswith("rwc")],
-    address_counter_groups=[group_name for group_name in group_map if group_name.startswith("adc")],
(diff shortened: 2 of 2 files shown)


adjordjevic-TT (commented):
If this is only done so tests are passing we should only edit testing part, if signals we are removing are generally uninformative this change is fine. Adding some LLK folks as reviewers since they are the ones that requested addition of these signals.

adjordjevic-TT:
@amahmudTT, @fvranicTT  can you guys take a look if you agree with removal of some RWC and ADC signals, if they have been useful to you (or you think they can be useful) we can just skip them in tests.

fvranicTT:
Adding @sstanisicTT who implemented that test originally.
For reference, we are talking about `z_state.reconfig.test_math_reconfig` test.

mcraigheadTT:
My read of the intent of this exalens code is: we have group_map and earlier in the file which declare the full list of all signals, vs. here we have register_window_counter_groups and address_counter_groups which declare the subset of signals which are specifically "RWC" and "ADC" state. The internal pipeline signals are still declared with names earlier in the file and people can still use them; we're just excluding them from the "RWC/ADC state" (which they are not).

That said, I am open to other ways of structuring the change as long as the outcome is the same.

sstanisicTT (approved):
IMO it makes complete sense to remove signals that point to internal/transient bits from this state.

adjordjevic-TT:
> My read of the intent of this exalens code is: we have group_map and earlier in the file which declare the full list of all signals, vs. here we have register_window_counter_groups and address_counter_groups which declare the subset of signals which are specifically "RWC" and "ADC" state. The internal pipeline signals are still declared with names earlier in the file and people can still use them; we're just excluding them from the "RWC/ADC state" (which they are not).
> 
> That said, I am open to other ways of structuring the change as long as the outcome is the same.

My only concern is this way we remove some signals from output of our `dump_tensix_state` command, since @sstanisicTT already agreed if @amahmudTT also agrees I have no problem with merging these changes.

mcraigheadTT:
Having dug through the RTL for a lot of these internal pipeline state signals - overall my impression is "good luck making sense of them unless you have the RTL in front of you." I would say they are probably most useful for advanced hang/race condition debug (though at that point personally I would probably be looking to switch to RTL anyway for the next level of debug). A typical end user is probably more interested in "give me the current values of all the ADCs, RWCs, SrcA/SrcB control state, etc." - the persistent architectural state.

Unfortunately, I also found that not all the ADCs are exposed, nor are the Blackhole implied src formats (!). There is persistent architectural state here that cannot be read.

adjordjevic-TT (approved):

mcraigheadTT:
Hi, it looks like this is approved but I don't have a "merge" button, could someone with appropriate permissions merge? Thank you!

PR #1106, opened 2026-07-02
Setting verbosity level in CI tests to DEBUG

### LINKED ISSUE #1105: Enable debug logs from umd for exalens app tests
Exalens app tests can fail and hang due to umd hanging, we are blind there as to what happened so it would be good to at least enable higher verbosity and umd logs for those tests.

adjordjevic-TT:
Closes #1105 

Setting verbosity when running app test in CI so we can better understand ND CI failures that we cannot easily reproduce.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -96,7 +96,7 @@ jobs:
       CONFIG: ci
       # So we can get all the makefile output we want
       VERBOSE: 5
-      LOGGER_LEVEL: INFO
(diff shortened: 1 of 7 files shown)


tt-vjovanovic:
@adjordjevic-TT beware that our test is reading app output and trying to verify with regex that everything works. Probably the easiest way is to redirect logging to error stream and just test standard output, but you can also filter out output inside the test that is logging (much harder and error prone approach).

tt-vjovanovic (approved):

PR #1107, opened 2026-07-02
Fix get_global/get_constant picking a shadowing type over the variable

onenezicTT:
## Problem
When a global shares its name with a type (e.g. `enum noc_mode` shadowing
`constexpr uint8_t noc_mode`), name resolution returned the type DIE (`DW_TAG_enumeration_type` instead of `DW_TAG_variable`), so
`get_global`/`get_constant` threw (`std::bad_cast`) or — for a folded
`constexpr` — read garbage (the const value was used as an address). This broke
tt-metal triage's `noc_mode` reads.

## Solution
- Resolve `get_global`/`get_constant` names to the variable, not a same-named type (across CUs).
- Only treat `DW_AT_const_value` as an address for pointers (not scalars).
- `get_global` now serves a folded constexpr's compile-time value (read-only) instead of reading device memory.

## Tests
Added an enum/variable name-collision fixture + regression test.

### DIFF
--- a/test/ttexalens/unit_tests/test_debug_symbols.py
+++ b/test/ttexalens/unit_tests/test_debug_symbols.py
@@ -10,7 +10,7 @@ from test.ttexalens.unit_tests.core_simulator import RiscvCoreSimulator
 from test.ttexalens.unit_tests.test_base import init_cached_test_context
 from ttexalens import OnChipCoordinate
 from ttexalens.context import Context
-from ttexalens.elf import DwarfDieTag, ElfFile, ElfVariable
+from ttexalens.elf import ElfFile, ElfVariable
 from ttexalens.exceptions import RiscHaltError
 from ttexalens.memory_access import MemoryAccess, create_memory_access
 from ttexalens.exceptions import RestrictedMemoryAccessError
@@ -392,10 +392,7 @@ class TestDebugSymbols(unittest.TestCase):
         self.assertEqual(42, self.parsed_elf.get_global("shadowed_const", TestDebugSymbols.mem_access).read_value())
         self.assertEqual(42, self.parsed_elf.get_constant("shadowed_const"))
 
-        die = self.parsed_elf.find_die_by_name("shadowed_const")
-        assert die is not None
-        self.assertEqual(DwarfDieTag.enumeration_type, die.tag)
-
+        # Enumerator lookup by qualified name keeps working
         self.assertEqual(1, self.parsed_elf.get_enum_value("shadowed_const::SHADOWED_B"))
 
     def test_elf_variable_array_iteration(self):
--- a/test/ttexalens/unit_tests/test_debug_symbols.py
+++ b/test/ttexalens/unit_tests/test_debug_symbols.py
@@ -10,7 +10,7 @@ from test.ttexalens.unit_tests.core_simulator import RiscvCoreSimulator
 from test.ttexalens.unit_tests.test_base import init_cached_test_context
 from ttexalens import OnChipCoordinate
 from ttexalens.context import Context
-from ttexalens.elf import ElfFile, ElfVariable
+from ttexalens.elf import DwarfDieTag, ElfFile, ElfVariable
 from ttexalens.exceptions import RiscHaltError
 from ttexalens.memory_access import MemoryAccess, create_memory_access
 from ttexalens.exceptions import RestrictedMemoryAccessError
@@ -385,6 +385,18 @@ class TestDebugSymbols(unittest.TestCase):
         self.assertEqual(-12345, self.parsed_elf.get_constant("c_int16_t"))
         self.assertEqual(-1234567, self.parsed_elf.get_constant("c_int32_t"))
         self.assertEqual(-1234567890123456789, self.parsed_elf.get_constant("c_int64_t"))
+
+    def test_elf_enum_variable_name_collision(self):
+        # `enum shadowed_const` and `constexpr uint8_t shadowed_const` share a
+        # name (mirrors dev_msgs.h's `enum noc_mode` shadowing `noc_mode`).
+        self.assertEqual(42, self.parsed_elf.get_global("shadowed_const", TestDebugSymbols.mem_access).read_value())
+        self.assertEqual(42, self.parsed_elf.get_constant("shadowed_const"))
+
+        die = self.parsed_elf.find_die_by_name("shadowed_const")
+        assert die is not None
+        self.assertEqual(DwarfDieTag.enumeration_type, die.tag)
+
+        self.assertEqual(1, self.parsed_elf.get_enum_value("shadowed_const::SHADOWED_B"))
 
     def test_elf_variable_array_iteration(self):
         variable_die = self.parsed_elf.find_die_by_name("g_global_struct")
--- a/ttexalens/cpp/elf/dwarf_die.cpp
+++ b/ttexalens/cpp/elf/dwarf_die.cpp
@@ -942,9 +942,11 @@ std::optional<ElfVariable> DwarfDie::read_value(const FrameInspection& frame) co
     }
 
     // Compile-time constant (DW_AT_const_value).
-    if (auto bytes = serialize_constant_value(get_constant_value(), resolved->get_size().value_or(4))) {
-        auto cache = std::make_shared<CachedReadMemoryAccess>(0, std::move(*bytes), NoMemoryAccess::instance());
-        return ElfVariable(resolved, 0, std::move(cache));
+    if (auto size = resolved->get_size()) {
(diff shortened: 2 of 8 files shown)


test/ttexalens/unit_tests/test_debug_symbols.py
```diff
@@ -386,6 +386,18 @@ def test_elf_variable_constants(self):
...
 
+    def test_elf_enum_variable_name_collision(self):
+        # `enum shadowed_const` and `constexpr uint8_t shadowed_const` share a
+        # name (mirrors dev_msgs.h's `enum noc_mode` shadowing `noc_mode`).
+        self.assertEqual(42, self.parsed_elf.get_global("shadowed_const", TestDebugSymbols.mem_access).read_value())
+        self.assertEqual(42, self.parsed_elf.get_constant("shadowed_const"))
+
+        die = self.parsed_elf.find_die_by_name("shadowed_const")
+        assert die is not None
+        self.assertEqual(DwarfDieTag.enumeration_type, die.tag)
```
tt-vjovanovic: Are you sure that will always be the case if you have two different DIEs with the same name?

ttexalens/cpp/elf/dwarf_die.cpp
```diff
@@ -811,7 +811,17 @@ static std::optional<uint64_t> get_address_recursed(const DwarfDie& die, bool al
             return std::nullopt;
         }
         if (const auto* cv = die.get_attribute(DwarfAttributeTag::const_value)) {
-            return attr_as_uint(cv);
+            // A const_value is only an *address* for the hard-coded-pointer
+            // pattern (`T* const g = (T*)0xADDR;`), where the constant IS the
+            // target address. For any non-pointer variable the const_value is
+            // the VALUE, not an address; treating it as one reads garbage from
+            // device memory (e.g. a folded `constexpr uint8_t x = 7` would read
+            // address 7). Such values are served via serialize_constant_value.
```
tt-vjovanovic: Consider rewriting comment in your own words (removing what you think is unnecessary LLM comment).

ttexalens/cpp/elf/dwarf_info.cpp
```diff
@@ -243,30 +289,44 @@ DwarfDie::ConstantValue DwarfInfo::get_constant(std::string_view name) const {
...
             }
         }
+        return ElfVariable(std::move(resolved_type), *address, std::move(memory_access));
+    }
+
+    // No runtime storage (folded constexpr): mask the difference by serving
+    // the compile-time DW_AT_const_value from a synthesized buffer. Read-only
+    // (writes raise via NoMemoryAccess), since there is nothing on device.
+    if (resolved_type) {
+        if (auto bytes = serialize_constant_value(die->get_constant_value(), resolved_type->get_size().value_or(4))) {
```
tt-vjovanovic: Is `4` what we want? Does `4` represent pointer size? Code should work on 32bit and 64bit CPUs

ttexalens/cpp/elf/dwarf_info.cpp
```diff
@@ -129,7 +130,40 @@ DwarfDiePtr DwarfInfo::find_function_by_address(uint64_t address) const {
...
+            return child;
+        }
+        if (!fallback) {  // copy, not move: `child` still advances the loop
+            fallback = child;
+        }
+    }
+    return fallback;
+}
+
+}  // namespace
```
tt-vjovanovic: I don't like that we are specializing search for variable and non-variable. When there is something else that we would like to search for you will add more specializations. I think it is easier to have `std::function<bool(const DwarfDiePtr&)> filter` argument that will filter out to what we are looking for (if it is `value->get_tag() == DwarfDieTag::variable` or something else) or `nullptr` if it is default (return first match as before). Adjust existing `find_child_by_name()` inside `DwarfDie` class to accept `std::function` argument instead of adding new private function.

test/ttexalens/unit_tests/test_debug_symbols.py
onenezicTT: Yeah this test relied on a specific case where we set the order (which is also not really guaranteed). Removed that check as it's not really the thing the test is interested in but rather the scenario of having same name with different types working

ttexalens/cpp/elf/dwarf_info.cpp
onenezicTT: Good callout

ttexalens/cpp/elf/dwarf_die.cpp
onenezicTT: Trimmed

ttexalens/cpp/elf/dwarf_info.cpp
onenezicTT: It was a fallback for the constant's own type size (only hit if DWARF has no size, which doesn't happen for a scalar constant). Removed it and removed the fallback in read_value as well.

ttexalens/cpp/elf/dwarf_die.hpp
```diff
@@ -368,4 +371,10 @@ class DwarfDie : public std::enable_shared_from_this<DwarfDie> {
     mutable std::optional<std::weak_ptr<DwarfDie>> parent;
 };
 
+// Serializes a DW_AT_const_value into `size` little-endian bytes suitable for
+// backing a synthesized ElfVariable (see ElfVariable::from_constant). Returns
+// std::nullopt for std::monostate. Bools become 0/1; floats/doubles keep their
+// IEEE-754 bit pattern.
+std::optional<std::vector<std::byte>> serialize_constant_value(const DwarfDie::ConstantValue& value, uint64_t size);
```
tt-vjovanovic: Consider moving this to be static function of `DwarfDie`.

ttexalens/cpp/elf/dwarf_info.cpp
```diff
@@ -144,15 +146,22 @@ DwarfDiePtr DwarfInfo::get_die_by_name(std::string_view name) const {
         start = pos + 2;
     }
 
+    // The filter constrains only the final component; intermediate scopes are
+    // matched by name alone.
```
tt-vjovanovic, line 150: I'm not sure if no filter is correct. Initially it started as getting variables from types, then it got extended to include enum values, then, types from namespaces and finally it got to be side effect of getting static local variables from functions. Maybe we should consider limiting what this functions should search for and have correct filter set for "parent" parts?

ttexalens/cpp/nanobind/bind_dwarf_die.cpp
```diff
@@ -235,8 +235,10 @@ void bind_dwarf_die(nb::module_& m) {
              nb::sig("def get_dereference_type(self) -> DwarfDie | None"))
         .def("get_array_element_type", &DwarfDie::get_array_element_type, nb::rv_policy::reference_internal,
              nb::sig("def get_array_element_type(self) -> DwarfDie | None"))
-        .def("find_child_by_name", &DwarfDie::find_child_by_name, nb::arg("name"), nb::rv_policy::reference_internal,
-             nb::sig("def find_child_by_name(self, name: str) -> DwarfDie | None"))
+        .def(
+            "find_child_by_name",
+            [](const DwarfDie& self, std::string_view name) { return self.find_child_by_name(name); }, nb::arg("name"),
+            nb::rv_policy::reference_internal, nb::sig("def find_child_by_name(self, name: str) -> DwarfDie | None"))
```
tt-vjovanovic: Why don't you expose lambda callback as well?

ttexalens/cpp/nanobind/bind_dwarf_info.cpp
```diff
@@ -20,8 +20,10 @@ namespace ttexalens::native_elf::bindings {
 void bind_dwarf_info(nb::module_& m) {
     nb::class_<DwarfInfo>(m, "DwarfInfo")
         .def("find_file_line_by_address", &DwarfInfo::find_file_line_by_address, nb::arg("address"))
-        .def("get_die_by_name", &DwarfInfo::get_die_by_name, nb::arg("name"), nb::rv_policy::reference_internal,
-             nb::sig("def get_die_by_name(self, name: str) -> DwarfDie | None"))
+        .def(
+            "get_die_by_name", [](const DwarfInfo& self, std::string_view name) { return self.get_die_by_name(name); },
+            nb::arg("name"), nb::rv_policy::reference_internal,
+            nb::sig("def get_die_by_name(self, name: str) -> DwarfDie | None"))
```
tt-vjovanovic: Why don't you expose lambda callback as well?

tt-vjovanovic (approved):

ttexalens/cpp/elf/dwarf_info.cpp
onenezicTT: Good concern, synced offline, will open an issue to track this and properly design.

PR #1112, opened 2026-07-13
Enable exalens starting with unhealthy devices

onenezicTT:
Umd now exposes the option to not crash if one of N devices are unhealthy for some reason during init. 
As step 1 - In exalens we now enable that option and warn about unhealthy devices. 

Follow up as Step 2 - enable triage to run in scenarios where one devices is unhealthy on a galaxy+.

Step 3 (TBD) - will enable triage and exalens to inspect those devices as UMD needs to expose some degraded device struct/view

### DIFF
--- a/ttexalens/umd_api.py
+++ b/ttexalens/umd_api.py
@@ -46,6 +46,16 @@ io_device_type: SIMULATION
 TLS_FOR_NOC_ID = threading.local()
 
 
+def _format_device_health(unhealthy_chips, health_errors: dict) -> str:
+    # An unhealthy chip may have no structured health error, so fall back to a placeholder.
+    parts = []
+    for chip_id in unhealthy_chips:
+        errors = health_errors.get(chip_id, [])
+        reason = ", ".join(type(error).__name__ for error in errors) if errors else "reason unavailable"
+        parts.append(f"chip {chip_id} ({reason})")
+    return "; ".join(parts)
+
+
 @Pyro5.api.expose
 class UmdApi:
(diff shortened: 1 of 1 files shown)


ttexalens/umd_api.py
```diff
@@ -132,6 +147,13 @@ def __init__(
             )
 
             if len(self.cluster_descriptor.get_all_chips()) == 0:
+                unhealthy_chips = self.cluster_descriptor.get_unhealthy_devices()
```
adjordjevic-TT: We can do this only once outside of this if and then use it here and afterwards when we need it.

ttexalens/umd_api.py
```diff
@@ -47,25 +46,40 @@ def create_simulation_cluster_descriptor(arch: tt_umd.ARCH) -> str:
 TLS_FOR_NOC_ID = threading.local()
 
 
+def _format_device_health(unhealthy_chips, health_errors: dict) -> str:
+    # An unhealthy chip may have no structured health error, so fall back to a placeholder.
+    parts = []
+    for chip_id in unhealthy_chips:
+        errors = health_errors.get(chip_id, [])
+        reason = ", ".join(type(error).__name__ for error in errors) if errors else "reason unavailable"
+        parts.append(f"chip {chip_id} ({reason})")
```
adjordjevic-TT: Maybe we want to call it device instead of chip since this is how we refer to it both in triage and exalens.

adjordjevic-TT (approved):

PR #1114, opened 2026-07-14
Fixing connecting to remote server on Quasar

adjordjevic-TT:
Since we changed initialization to be NOC1 by default we have hack that switches to NOC0 if it is a simulation so Quasar works. We need to do same thing when trying to connect to exalens server or we would end up with noc mismatch.

Also since SMN is enumerated as 2 we are getting out of bounds error when accessing device coords cache so now we are making list of 3 elements with 1 None for Quasar. Other option is to switch to dictionary which would be slower.


ttexalens/tt_exalens_init.py
```diff
@@ -79,6 +79,11 @@ def load_context(
     safe_mode: bool = True,
 ) -> Context:
     """Load the TTExaLens context object with specified parameters."""
+
+    # Since Quasar is only available through simulation as workaround if we are using simulator we switch to NOC0
+    if umd_api.simulation_directory is not None and noc_id == NocId.NOC1:
```
tt-vjovanovic: Instead of testing `noc_id == NocId.NOC1` consider allowing `noc_id` to be `None`. Then you can test if value has been sent at all (which in simulation it probably won't be changed). This causes user to worry about this instead of the library.

tt-vjovanovic: On the other hand, this fix seems like you have problem with `init_ttexalens_remote` as it provides `noc_id` as default value (which is `NOC1`). Consider fixing it by using remote context `noc_id`. You can store `initialization_noc_id` field inside `UmdApi` that will be what was used in constructor.

ttexalens/tt_exalens_init.py
```diff
@@ -74,11 +74,14 @@ def init_ttexalens_remote(
...
     file_api: FileAccessApi,
-    noc_id: NocId = NocId.NOC1,
+    noc_id: NocId | None = None,
     noc_failover: bool = True,
     safe_mode: bool = True,
 ) -> Context:
     """Load the TTExaLens context object with specified parameters."""
+    if noc_id is None:
+        noc_id = umd_api.initialization_noc_id
+
```
tt-vjovanovic: Consider moving this to line 70 without `if`.

ttexalens/tt_exalens_init.py
```diff
@@ -74,11 +74,14 @@ def init_ttexalens_remote(
 def load_context(
     umd_api: UmdApi,
     file_api: FileAccessApi,
-    noc_id: NocId = NocId.NOC1,
+    noc_id: NocId | None = None,
```
tt-vjovanovic: Remove default value from `noc_id` parameter. After providing this argument from `init_ttexalens_remote`, all places will supply value and default value is not needed.

tt-vjovanovic (approved):

PR #1117, opened 2026-07-16
Adding read and write memory bytes to rocket core debug

adjordjevic-TT:
Adding read and write memory bytes to `RocketCoreDebug` the same way it is implemented in `BabyRiscDebug`. This is needed for full callstack to work.


miacim:
@adjordjevic-TT it would be good if you add some logging at least at log level debug.

miacim (approved):

ttexalens/hardware/rocket_core_debug.py
```diff
@@ -65,19 +75,74 @@ def write_gpr(self, register_index: int, value: int) -> None:
...
+            suffix_size = aligned_end - (address + size_bytes)
+            suffix_word = self._read_word(aligned_end - word_size)
+            new_data.extend(suffix_word.to_bytes(word_size, byteorder="little")[-suffix_size:])
+
+        assert len(new_data) % word_size == 0, "Data length must be multiple of word size after alignment"
+
+        for offset in range(0, len(new_data), word_size):
+            new_addr = aligned_start + offset
+            word = int.from_bytes(new_data[offset : offset + word_size], byteorder="little")
+            self._write_word(new_addr, word)
```
tt-vjovanovic: This code looks exactly the same as in `baby_risc_debug.py`. Consider having utility method that would be reused on both places. Ideally, it would be the same one we had when we used 4 byte mode in `umd_device.py`. This one has unnecessary double allocation. You would allocated new array in order to send 4 byte `int` values to other method that knows how to write word. Using `memoryview` would remove double allocation and all you need is handling first unaligned word and last unaligned word (that all existed in `umd_device.py`, just there we didn't send words, but "subarrays" of 4 bytes).

ttexalens/hardware/rocket_core_debug.py
```diff
@@ -65,19 +75,74 @@ def write_gpr(self, register_index: int, value: int) -> None:
...
+        pos = 0
+        while pos < size_bytes:
+            addr = address + pos
+            word_addr = addr - (addr % word_size)
+            word = self._read_word(word_addr)
+            word_bytes = word.to_bytes(word_size, byteorder="little")
+            start_in_word = addr - word_addr
+            n = min(word_size - start_in_word, size_bytes - pos)
+            buffer[pos : pos + n] = word_bytes[start_in_word : start_in_word + n]
+            pos += n
```
tt-vjovanovic: This code looks exactly the same as in baby_risc_debug.py. Consider having utility method that would be reused on both places.

tt-vjovanovic (approved):

PR #1129, opened 2026-07-20
Implementing step fucntionality for rocket cores

adjordjevic-TT:
Implemented single step functionality for rocket cores.

Added function for executing program buffer and used it in few places we did it manually.

Implemented read/write CSR and enabled writing to PC register with write GPR.

### DIFF
--- a/ttexalens/hardware/quasar/rocket_core_debug.py
+++ b/ttexalens/hardware/quasar/rocket_core_debug.py
@@ -29,7 +29,6 @@ INSN_CSRR_X5_DPC = 0x7B1022F3  # csrr x5, dpc  (csrrs x5, 0x7B1, x0)
 INSN_EBREAK = 0x00100073
 # DCSR (CSR 0x7B0) single-step control. dcsr.step is set/cleared atomically with a
 # single csrrsi/csrrci from the program buffer
-DCSR_STEP = 1 << 2
 INSN_CSRSI_DCSR_STEP = 0x7B026073  # set dcsr.step
 INSN_CSRCI_DCSR_STEP = 0x7B027073  # clear dcsr.step
 INSN_CSRW_DPC_X5 = 0x7B129073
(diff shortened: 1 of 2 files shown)


ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -27,8 +27,18 @@
 
 INSN_CSRR_X5_DPC = 0x7B1022F3  # csrr x5, dpc  (csrrs x5, 0x7B1, x0)
 INSN_EBREAK = 0x00100073
+# DCSR (CSR 0x7B0) single-step control. dcsr.step is set/cleared atomically with a
+# single csrrsi/csrrci from the program buffer
+DCSR_STEP = 1 << 2
+INSN_CSRSI_DCSR_STEP = 0x7B026073  # set dcsr.step
+INSN_CSRCI_DCSR_STEP = 0x7B027073  # clear dcsr.step
+INSN_CSRW_DPC_X5 = 0x7B129073
```
tt-vjovanovic:
```suggestion
# Debugging instructions
INSN_CSRR_X5_DPC = 0x7B1022F3  # csrrs x5, dpc, x0
INSN_EBREAK = 0x00100073 # ebreak
INSN_CSRSI_DCSR_STEP = 0x7B026073  # csrrsi x0, dcsr, 4
INSN_CSRCI_DCSR_STEP = 0x7B027073  # csrrci x0, dcsr, 4
INSN_CSRW_DPC_X5 = 0x7B129073 # csrrw x0, dpc, x5
```

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -27,8 +27,18 @@
 
 INSN_CSRR_X5_DPC = 0x7B1022F3  # csrr x5, dpc  (csrrs x5, 0x7B1, x0)
 INSN_EBREAK = 0x00100073
+# DCSR (CSR 0x7B0) single-step control. dcsr.step is set/cleared atomically with a
+# single csrrsi/csrrci from the program buffer
+DCSR_STEP = 1 << 2
```
tt-vjovanovic: `DCSR_STEP` doesn't seem to be used in this PR

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -189,6 +231,15 @@ def _execute_abstract_command(self, command: int, prepare: Callable[[], None] |
...
+
+        def stage_progbuf() -> None:
+            self.register_store.write_register("TT_DEBUG_MODULE_APB_PROGBUF0", insn0)
+            self.register_store.write_register("TT_DEBUG_MODULE_APB_PROGBUF1", insn1)
+
+        self._execute_abstract_command(CMD_EXEC_PROGBUF, stage_progbuf)
+
     def _read_gpr_via_debug_module(self, index: int) -> int:
         """Read a 64-bit general purpose register via the Access Register abstract command."""
         with self.ensure_debug_module_is_active():
```
tt-vjovanovic: Is is possible to read GPR if core is not in halted state? Can it be in halted state if debug module is not active?

ttexalens/hardware/quasar/rocket_core_debug.py
```diff
@@ -206,16 +257,17 @@ def stage_value() -> None:
 
         self._execute_abstract_command(CMD_WRITE_GPR_BASE + index, stage_value)
 
+    def _read_csr_through_debug_module(self, read_csr_insn: int) -> int:
+        self._execute_program_buffer(read_csr_insn, INSN_EBREAK)
+        return self._read_gpr_via_debug_module(SCRATCH_GPR_INDEX)
```
tt-vjovanovic, line 250: This code does two "transactions". If you take a look at method implementation, `_execute_program_buffer` will ensure that core has debug module active and then that it is halted, while `_read_gpr_via_debug_module` will ensure that debug module is active afterwards. So unrolled code looks something like this:
```python
with self.ensure_debug_module_is_active():
    with self.ensure_halted():
        # Do something
with self.ensure_debug_module_is_active():
    # Do something else
```
Is it required that core is halted during execution of this method? Please go through methods of this class and do `ensure_...` only for self contained methods (like `get_pc()`). For others you should throw if requirement is not satisfied. Otherwise, you will end up having many calls to `ensure_` something and then returning into original state...

ttexalens/hardware/quasar/rocket_core_debug.py
adjordjevic-TT: It is not possible since executing abstract commands (which is used here) requires core to be halted. If debug module is not active we cannot perform the check if core is halted so I would say no.

ttexalens/hardware/quasar/rocket_core_debug.py
adjordjevic-TT: I ended up changing it to not put debug module back to reset and just take it out instead if it is in reset.

tt-vjovanovic (approved):

Issue #960, opened 2026-03-27
Dumping overlay block's state


adjordjevic-TT:
We need to implement dumping overlay block's state (similar what we do with tensix). Investigate what registers are useful, document them and expose them so we can read from or write to them. Metal is already exposing these in [here](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/inc/internal/tt-2xx/quasar/overlay/meta/registers) so it is a good starting point. Also we should see what else is important (ex. command buffer).

PR #1137, opened 2026-07-27
Adding dump overlay state

adjordjevic-TT:
Closing #960 

Added `dump-overlay-state` command which dumps several groups of register containing useful information for rocket core debugging.

Added missing registers to `regsiter_map`.

Added memory blocks containing registers to overlay block.

Added overlay register descriptions.


ttexalens/cli_commands/dump-rocket-state.py
```diff
@@ -0,0 +1,303 @@
...
+
+
+def _dec(v: int | None) -> str:
+    return "-" if v is None else str(v)
+
+
+def _flag(v: int | None) -> str:
+    if v is None:
+        return "-"
+    return "False" if v == 0 else "True"
```
tt-vjovanovic: These seems like utility methods that should be probably shared with other commands. Either other scripts have similar, or should share these ones.

ttexalens/hardware/quasar/functional_overlay_registers.py
```diff
@@ -1287,4 +2589,299 @@ class NeoRegisterDescription(RegisterDescription):
...
+    "TT_CLUSTER_CLINT_MSIP_6_": ClintRegisterDescription(offset=0x18),
+    "TT_CLUSTER_CLINT_MSIP_7_": ClintRegisterDescription(offset=0x1C),
+    "TT_CLUSTER_CLINT_MTIMECMP_0_": ClintRegisterDescription(offset=0x4000),
+    "TT_CLUSTER_CLINT_MTIMECMP_1_": ClintRegisterDescription(offset=0x4008),
+    "TT_CLUSTER_CLINT_MTIMECMP_2_": ClintRegisterDescription(offset=0x4010),
+    "TT_CLUSTER_CLINT_MTIMECMP_3_": ClintRegisterDescription(offset=0x4018),
+    "TT_CLUSTER_CLINT_MTIMECMP_4_": ClintRegisterDescription(offset=0x4020),
+    "TT_CLUSTER_CLINT_MTIMECMP_5_": ClintRegisterDescription(offset=0x4028),
+    "TT_CLUSTER_CLINT_MTIMECMP_6_": ClintRegisterDescription(offset=0x4030),
+    "TT_CLUSTER_CLINT_MTIMECMP_7_": ClintRegisterDescription(offset=0x4038),
```
tt-vjovanovic, line 2755: Some register names end with `_`. Is this intended?

ttexalens/cli_commands/dump-overlay-state.py
```diff
@@ -0,0 +1,303 @@
...
+Options:
+  --counters        Dump LLK tile counters (DFB credits / flow control).
+  --cmdbuf          Dump ROCC command-buffer descriptors (in-flight NOC transfers).
+  --errors          Dump bus error units (latched fault cause / address).
+  --wdt             Dump watchdog timers (hang detection).
+  --debug           Dump RISC-V debug module state (halt / abstract / system bus).
+  --clint           Dump CLINT (per-core software interrupt + timer compare).
+  --plic            Dump PLIC (interrupt priorities / pending / enables).
+  --cores <cores>   Comma-separated Rocket core indices for the per-core groups (e.g. 0,3,5). Default: all cores.
+
```
tt-vjovanovic, line 17: Are you sure that all of these are related to Rocket cores (I'm assuming from the name that `dump-rocket-state` means Rocker cores)? If not, maybe script name is incorrect, or it should be split into multiple scripts. If something is reorganized in new hardware, will this script still work?

ttexalens/cli_commands/dump-rocket-state.py
```diff
@@ -0,0 +1,303 @@
+# SPDX-FileCopyrightText: © 2025 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+"""
+Usage:
+  dump-rocket-state [--counters] [--cmdbuf] [--errors] [--wdt] [--debug] [--clint] [--plic] [--cores <cores>] [ -d <device> ] [ -l <loc> ] [ -v ]
```
tt-vjovanovic: `dump-tensix-state` script relies on `Device` class returning maps that can help it print necessary data. Why this script is not made like that?

ttexalens/cli_commands/dump-rocket-state.py
```diff
@@ -0,0 +1,303 @@
...
+  dump-rocket-state                 # all groups, current device/core
+  rocket                            # same (short name)
+  rocket --cmdbuf                   # only ROCC command buffers
+  rocket --errors --wdt             # bus errors + watchdogs together
+  rocket --counters -v              # LLK counters incl. read-mirrors/thresholds
+  rocket --cmdbuf --cores 0,3,5     # command buffers for cores 0, 3, 5 only
+  rocket --debug -l 1,1             # debug module at location 1,1
+"""
+
+import tabulate
```
tt-vjovanovic: New scripts should use rich tables. We have an issue to replace all scripts to use rich as well.

ttexalens/hardware/quasar/functional_overlay_registers.py
adjordjevic-TT: These are copied from `tt-metal` where they end with `_`.

ttexalens/cli_commands/dump-overlay-state.py
adjordjevic-TT: These are related to overlay block, I was debating between naming command `dump-rocket-state` and `dump-overlay-state` and ultimately decided for the first because I was thinking `overlay` might make people think about noc overlay on wormhole/blackhole. What do you think about this?

adjordjevic-TT: `dump-overlay-state` is for sure more accurate

tt-vjovanovic: Let's go with `dump-overlay-state`, but we don't need to support old Overlay from GS, WH, BH... We can add it later if someone needs it.

tt-vjovanovic (approved):

PR #1147, opened 2026-07-30
Keep QSR cores in reset during simulator init

mcraigheadTT:
Revert the QSR branch added to tt-exalens simulator initialization. That branch deasserted RiscType.ALL for every QSR Tensix core during generic context setup, but no reset-vector stub or DM/RV64 ELF is loaded at that point. On QSR this can wake DM/RV64 cores and start them from junk reset contents before the LLK harness has installed code, which is unsafe and causes simulator instruction decode failures. Keep the existing Blackhole workaround, where a BRISC infinite-loop stub is written before release, and leave QSR cores in reset until a caller explicitly loads and starts the specific cores it needs.

### DIFF
--- a/ttexalens/umd_api.py
+++ b/ttexalens/umd_api.py
@@ -110,13 +110,11 @@ class UmdApi:
             # and take all cores out of reset. Downstream test harnesses then re-assert specific
             # cores, load their ELFs, and re-deassert. This keeps cores from executing garbage
             # between tt-exalens init and the harness taking over.
-            for core in soc_descriptor.get_cores(tt_umd.CoreType.TENSIX):
-                core_noc0 = soc_descriptor.translate_coord_to(core, tt_umd.CoordSystem.NOC0)
-                if tt_device.get_arch() == tt_umd.ARCH.BLACKHOLE:
+            if tt_device.get_arch() == tt_umd.ARCH.BLACKHOLE:
+                for core in soc_descriptor.get_cores(tt_umd.CoreType.TENSIX):
+                    core_noc0 = soc_descriptor.translate_coord_to(core, tt_umd.CoordSystem.NOC0)
                     tt_device.noc_write32(core_noc0.x, core_noc0.y, 0, 0x6F)
                     tt_device.deassert_risc_reset(tt_umd.tt_xy_pair(core.x, core.y), tt_umd.RiscType.BRISC)
-                elif tt_device.get_arch() == tt_umd.ARCH.QUASAR:
-                    tt_device.deassert_risc_reset(tt_umd.tt_xy_pair(core.x, core.y), tt_umd.RiscType.ALL)
             cluster_descriptor_content = create_simulation_cluster_descriptor(tt_device.get_arch())
             self.cluster_descriptor = tt_umd.ClusterDescriptor.create_from_yaml_content(cluster_descriptor_content)
             self.devices[0] = UmdDevice(


ttexalens/umd_api.py
```diff
@@ -110,13 +110,11 @@ def __init__(
...
-            for core in soc_descriptor.get_cores(tt_umd.CoreType.TENSIX):
-                core_noc0 = soc_descriptor.translate_coord_to(core, tt_umd.CoordSystem.NOC0)
-                if tt_device.get_arch() == tt_umd.ARCH.BLACKHOLE:
+            if tt_device.get_arch() == tt_umd.ARCH.BLACKHOLE:
+                for core in soc_descriptor.get_cores(tt_umd.CoreType.TENSIX):
+                    core_noc0 = soc_descriptor.translate_coord_to(core, tt_umd.CoordSystem.NOC0)
                     tt_device.noc_write32(core_noc0.x, core_noc0.y, 0, 0x6F)
                     tt_device.deassert_risc_reset(tt_umd.tt_xy_pair(core.x, core.y), tt_umd.RiscType.BRISC)
-                elif tt_device.get_arch() == tt_umd.ARCH.QUASAR:
-                    tt_device.deassert_risc_reset(tt_umd.tt_xy_pair(core.x, core.y), tt_umd.RiscType.ALL)
```
tt-vjovanovic, line 119: You need to verify that Quasar simulator and Emulator work after this change.
This "fix" was added because RTL simulator has "feature" that requires deasserting before continuing, or deassert through debug registers won't work.

Take a look at https://github.com/tenstorrent/tt-umd/blob/db90f47eb3668cc8afafd39a9d92e2d85eaf8f20/device/tt_device/rtl_simulation_tt_device.cpp#L256

tt-vjovanovic: LLK test infra had problem with this, so we needed to fix it for Quasar. Here are links to tt-exalens guides for:
- [Quasar emulation](https://yyz-gitlab.local.tenstorrent.com/tensix/tt-umd-simulators/-/blob/main/emu/tt-exalens.md?ref_type=heads)
- [Quasar RTL simulator](https://yyz-gitlab.local.tenstorrent.com/tensix/tt-umd-simulators/-/blob/main/vcs/tt-exalens.md?ref_type=heads)

For LLK test infra, @fvranicTT can help with guide.

mcraigheadTT: If this requires additional testing or additional fixes, I think that work needs to be done by the LLK or exalens team, not by the simulator team. This is currently gating our ability to run any LLK tests at all our of the box, and the code appears to be clearly wrong: it appears to be taking these cores out of reset into uninitialized memory, which has unboundedly-bad consequences. Additionally, I don't think we want to run the DM core during LLK tests; at best, this would slow down every single simulation.

As this is causing 100% of tests to fail, an exalens fix here is high priority for our QSR ttsim work.

tt-vjovanovic: Without verification this change cannot be merged as it will break existing scenarios with RTL simulators/emulators.
If it is needed, we can always put `0x6f` on start addresses, so that all riscs that are deasserted execute endless while loop, but reverting deassert is not going to work as expected.

mcraigheadTT: An infinite loop is better than uninitialized memory, but it is definitely not good and will slow down every simulation.

tt-vjovanovic: Other option that we can do is expose information if it is RTL simulation/emulation from UMD. Currently we are not aware which simulator is being used in tt-exalens.

mcraigheadTT:
We will definitely need a solution for this soon, as this is now the only thing preventing at least one QSR LLK test from passing on the latest ttsim release.

ttexalens/umd_api.py
tt-vjovanovic: Actually, few lines above in this PR is logic that chooses simulation type.
This PR: https://github.com/tenstorrent/tt-exalens/pull/1155
fixes problem and speeds up ttsim path.

tt-vjovanovic:
Closing this in favor of https://github.com/tenstorrent/tt-exalens/pull/1155

mcraigheadTT:
Thanks! The other one looks like it should work, but we will verify

PR #1153, opened 2026-08-10
Elf variable reference fixes

onenezicTT:
- References are now walked like the object they name (read/write/dereference/etc).
- A pointer/reference held in a register (DW_OP_reg*) now falls back to the frame's memory access instead of NoMemoryAccess, so it can actually be followed.
- Fixes exception translation: `bindings.cpp` was missing nanobind's std::string/std::optional casters, so every native error surfaced as `RuntimeError: std::bad_cast` instead of `SymbolNotFoundError`/`TypeMismatchError`/`DataLossError`/etc.

### DIFF
--- a/riscv-src/callstack.cc
+++ b/riscv-src/callstack.cc
@@ -277,6 +277,11 @@ __attribute__((noinline)) void dispatch(uint32_t type_index) {
     register type reg_value asm("s2") = host_value<type>((slot) * 8); \
     asm volatile("" : "+r"(reg_value))
 
+// As above, but binds a reference to that slot, so s2 holds the referent's address.
+#define DECLARE_CALLEE_SAVED_REFERENCE(type, slot)                                                      \
+    register type& reg_value asm("s2") = *reinterpret_cast<type*>(mailbox_value_buffer() + (slot) * 8); \
+    asm volatile("" : : "r"(&reg_value))
+
 // Keeps reg_value live past the nested call so it is preserved in - or spilled from - s2.
 #define USE_CALLEE_SAVED_VALUE() asm volatile("" : : "r"(reg_value))
 
@@ -361,7 +366,19 @@ __attribute__((noinline, noclone)) void value_test_charptr() {
 }
 }  // namespace callee_saved_test
 
+namespace reference_test {
+// Like callee_saved_test, but s2 holds a reference. Its register-direct DWARF
+// location makes it materialised, so following it has to reach live memory.
+
+__attribute__((noinline, noclone)) void run() {
+    DECLARE_CALLEE_SAVED_REFERENCE(volatile uint32_t, 0);
+    halt();
+    USE_CALLEE_SAVED_VALUE();
+}
+}  // namespace reference_test
+
 #undef DECLARE_CALLEE_SAVED_VALUE
+#undef DECLARE_CALLEE_SAVED_REFERENCE
 #undef USE_CALLEE_SAVED_VALUE
 
 namespace tail_call_test {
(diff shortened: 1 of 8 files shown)


riscv-src/callstack.cc
```diff
@@ -522,7 +539,8 @@ __attribute__((noinline)) int run(int a) {
 int main() {
     if (*g_MAILBOX != 0xFFFFFFFF && *g_MAILBOX != 0xFFFFFFFE && *g_MAILBOX != 0xFFFFFFFD && *g_MAILBOX != 0xFFFFFFFC &&
         *g_MAILBOX != 0xFFFFFFFB && *g_MAILBOX != 0xFFFFFFFA && *g_MAILBOX != 0xFFFFFFF9 && *g_MAILBOX != 0xFFFFFFF8 &&
-        *g_MAILBOX != 0xFFFFFFF7 && (*g_MAILBOX & 0xFFFFFF00) != 0xFFFFFE00 && (*g_MAILBOX < 0 || *g_MAILBOX > 1000)) {
+        *g_MAILBOX != 0xFFFFFFF7 && *g_MAILBOX != 0xFFFFFFF6 && (*g_MAILBOX & 0xFFFFFF00) != 0xFFFFFE00 &&
+        (*g_MAILBOX < 0 || *g_MAILBOX > 1000)) {
         *g_MAILBOX = 10;
     }
```
tt-vjovanovic: Consider making this easier to read.

ttexalens/cpp/elf/variable.cpp
```diff
@@ -213,6 +215,9 @@ ElfVariable ElfVariable::dereference() const {
 
 ElfVariable ElfVariable::get_index(int64_t i) const {
     const auto tag = type_die->get_tag();
+    if (is_reference_tag(tag)) {
+        return dereference().get_index(i);
+    }
```
tt-vjovanovic, line 220: This doesn't make sense. You cannot index references.

ttexalens/cpp/elf/variable.cpp
```diff
@@ -237,7 +242,11 @@ ElfVariable ElfVariable::get_index(int64_t i) const {
 }
 
 uint64_t ElfVariable::get_length() const {
-    if (type_die->get_tag() != DwarfDieTag::array_type) {
+    const auto tag = type_die->get_tag();
+    if (is_reference_tag(tag)) {
+        return dereference().get_length();
+    }
```
tt-vjovanovic, line 248: Hm... Similar to how length is implemented (reference is treated as it is not a pointer, but it actually is), maybe we should revisit all operations on top of reference? For example:
```
int a = 5, b = 10;
int &r = a;

if (r < b) return r;
return r *  b;
```
This code is valid, so maybe we should implement same functionality for `ElfVariable` as well. Thoughts?

ttexalens/cpp/elf/variable.cpp
```diff
@@ -351,6 +364,12 @@ void ElfVariable::write_value(Value value, bool check_data_loss) {
     auto type = type_die;
     const auto tag = type->get_tag();
 
+    if (is_reference_tag(tag)) {
```
tt-vjovanovic, line 367: I'm not sure what makes sense for rvalue reference or for const reference...

tt-vjovanovic (approved):

ttexalens/cpp/elf/variable.cpp
onenezicTT: If you have a ref to an array, why wouldn't you be able to index it? This is exactly one of the cases that are tested in the tests

tt-vjovanovic: My bad. I didn't see `dereference()` part of on line 219. Eye exam is still waiting for me :D

ttexalens/cpp/elf/variable.cpp
onenezicTT: Left them untouched for now, we talked a bit offline on what can be done and will open an issue just to track.

ttexalens/cpp/elf/variable.cpp
onenezicTT: All of those operators are already covered by this change, I added some more tests just to validate they are actually covered.

PR #1155, opened 2026-08-18
Apply fix for simulation only for RTL simulation

tt-vjovanovic:
ttsim doesn't need this fix and fix is making it run slower.


onenezicTT (approved):

fvranicTT (approved):
I tested this fix locally on ttsim and emulator - it works well. I'd even say that ttsim is faster than it was before.

PR #1165, opened 2026-08-26
Introducing NEO ID to CLI

### LINKED ISSUE #1140: Add neo id support to CLI
Currently in CLI users are unable to specify `neo_id` which makes some of the commands that use API that supports `neo_id` crash. We should add `neo_id` to `UIState` and possibly as arguments to commands where needed. `neo_id=None` would select overlay block, and `neo_id=1...4` would select corresponding NEO.

adjordjevic-TT:
Closes #1140 

For commands that target tensix to work on Quasar we need to specify NEO ID.

Added NEO ID to `UIState` and also go command can now switch between NEOs and overlay block, in output neo is only displayed if block has it, otherwise it is left out.

Added neo_id arguments to commands as well and passed it to API that takes it as argument.

We specify NEO with 0..3 and overlay block with None.


ttexalens/cli_commands/callstack.py
```diff
@@ -18,12 +18,14 @@
 
 Examples:
   callstack build/riscv-src/wormhole/sample.brisc.elf -r brisc
+  callstack sample.trisc0.elf -r trisc0 --neo 1
```
tt-vjovanovic: Update sample paths: `build/riscv-src/wormhole/sample.debug.brisc.elf`.

ttexalens/hardware/quasar/functional_worker_block.py
```diff
@@ -52,63 +52,51 @@ def __init__(self, location: OnChipCoordinate):
...
-            + self.neo3.noc_memory_list
+            + [memory for neo in self.neos for memory in neo.noc_memory_list]
             + self.overlay.noc_memory_list
         )
 
+    @property
+    def neo_ids(self) -> list[int]:
+        return [neo.neo_id for neo in self.neos]
+
+    def get_neo_by_id(self, neo_id: int | None) -> QuasarFunctionalNeoBlock | QuasarFunctionalOverlayBlock | None:
```
tt-vjovanovic: Consider making better name for this method since it also returns `QuasarFunctionalOverlayBlock`

ttexalens/hardware/noc_block.py
```diff
@@ -85,3 +92,36 @@ def get_risc_debug(self, risc_name: str, neo_id: int | None = None) -> RiscDebug
         This method should be overridden in subclasses to provide a specific implementation.
         """
         raise NotImplementedError(f"Noc block on location {self.location.to_user_str()} doesn't have RISC cores.")
+
+
+# The name users type to select the overlay block instead of one of the NEOs.
+NEO_ID_OVERLAY_NAME = "overlay"
```
tt-vjovanovic: This should be part of naming update as well. `NEO_ID = overlay` sounds a bit strange...

tt-vjovanovic: Omitting one makes sense

ttexalens/uistate.py
```diff
@@ -83,6 +86,25 @@ def __init__(self, context: Context) -> None:
...
+
+    @property
+    def has_neos(self) -> bool:
+        return len(self.current_neo_ids) > 0
+
+    def set_current_neo_id(self, neo_id: int | None) -> None:
+        if neo_id is not None and neo_id not in self.current_neo_ids:
+            util.WARN(
+                f"Invalid NEO {neo_id} for the block at {self.current_location.to_user_str()}. Setting NEO ID skipped."
+            )
```
tt-vjovanovic, line 105: Even though setting NEO ID says it will be skipped, it isn't and it will be saved, so either add `return` or `else`.

tt-vjovanovic (approved):

ttexalens/hardware/quasar/functional_worker_block.py
adjordjevic-TT: renamed to `get_sub_block`.

ttexalens/hardware/noc_block.py
adjordjevic-TT: Changed to None.

PR #1166, opened 2026-09-01
Adding Quasar tensix configuration registers

### LINKED ISSUE #1023: Add tensix configuration registers for Quasar
We should add configuration registers for tensix on quasar. They can be found here https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/inc/internal/tt-2xx/quasar/cfg_defines.h. 

Base addresses per NEO are `0x00800000`, `0x00810000`, `0x00820000`, `0x00830000`.
Each TRISC can also access their own configuration registers.

We also need to add `TensixRegisterDescriptions` grouping configuration registers into logical parts.

adjordjevic-TT:
Closes #1023

Changed `get_default_risc_debug` to accept `neo_id` as argument with default being `None`. Earlier archs ignore this argument, but functional worker on Quasar uses it to specify NEO from which we want to get default risc.

Added list of configuration registers found in https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/inc/internal/tt-2xx/quasar/cfg_defines.h. Also, grouped registers into logical groups like we have for wh/bh.

Changed private address offset for configuration registers from `0xA000` to `0x20000`, evidence: https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/inc/internal/tt-2xx/quasar/tt_t6_trisc_map.h, will run a llk test that sets config registers to confirm this is now correct. Also, removed noc address since config registers are not accessible directly from NoC.


ttexalens/hardware/quasar/functional_neo_block.py
```diff
@@ -38,10 +39,7 @@ def get_register_base_address(register_description: RegisterDescription) -> Devi
         assert neo_base_address.noc_address is not None, "RISC base address must have a NOC address"
 
         if isinstance(register_description, ConfigurationRegisterDescription):
-            return DeviceAddress(
-                private_address=neo_base_address.private_address + 0xA000,
-                noc_address=neo_base_address.noc_address + 0xA000,
-            )
+            return DeviceAddress(private_address=neo_base_address.private_address + 0x20000)
```
tt-vjovanovic, line 42: Does this change means that we cannot access them over NOC?

ttexalens/hardware/quasar/functional_neo_block.py
```diff
@@ -209,6 +236,9 @@ def all_riscs(self) -> list[RiscDebug]:
             self.get_risc_debug(self.trisc3.risc_name),
         ]
 
+    def get_default_risc_debug(self) -> RiscDebug:
+        return self.get_risc_debug(self.trisc0.risc_name)
```
tt-vjovanovic: Maybe we should use `trisc3` as default one?

ttexalens/hardware/quasar/tensix_register_description.py
```diff
@@ -0,0 +1,88 @@
...
+        ),
+        src_config=_group(
+            r"(?P<block>CLR_DVALID|FIDELITY_BASE|DISABLE_IMPLIED_SRCA_FMT|DISABLE_IMPLIED_SRCB_FMT)"
+            r"_SEC(?P<instance>\d)_(?P<field>.+)",
+            NUM_CONFIG_SECTIONS,
+        ),
+        sfpu_config=_group(
+            r"(?P<block>SFPU_DEST_FMT|SFPU_STACK)_SEC(?P<instance>\d)_(?P<field>.+)", NUM_CONFIG_SECTIONS
+        ),
+    )
```
tt-vjovanovic: To be honest, I'm not sure this is the right way to go. Maintaining those regex seems imposible for human to debug. Since it is single pass to do it, why don't we have them explicit like in other architectures?

tt-vjovanovic (approved):

ttexalens/hardware/quasar/functional_neo_block.py
adjordjevic-TT: Yeah, it is not possible to directly access configuration registers via NoC therefore I removed `noc_address`.

ttexalens/hardware/quasar/tensix_register_description.py
adjordjevic-TT: Done. Also, now base class `TensixRegisterDescription` is empty and each architecture must define its own groups.

Issue #1142, opened 2026-07-29
Enable telemetry on Quasar


adjordjevic-TT:
On Quasar, there is no ARC core, which is typically used to read postcode and telemetry data (e.g., heartbeat, clock frequency, temperature). Instead, this functionality can be accessed through the SMC. The postcode is exposed via a scratch register at address `0xC0010108`. 

The telemetry table follows the same structure as on Wormhole and Blackhole, with values located in the SMC’s L1 memory. This should be available through SMN.

adjordjevic-TT:
This should be provided by UMD and once it is we should just harness it.

adjordjevic-TT:
We switched to UMD api for firmware telemetry reading. This will automatically support Quasar once telemetry is implemented by firmware and exposed to us through UMD.

Currently, this issue is blocked by https://github.com/tenstorrent/tt-umd/issues/2501.

PR #1168, opened 2026-09-02
Use umd's telemetry tags instead of hardcoded list

onenezicTT:
In `arc_block` we hardcoded a list of telemetry tags, which since then changed and updated a couple times in umd. Now umd exposes that map so we should align.

### DIFF
--- a/ttexalens/hardware/arc_block.py
+++ b/ttexalens/hardware/arc_block.py
@@ -18,49 +18,8 @@ from ttexalens.util import FirmwareVersion
 
 CUTOFF_FIRMWARE_VERSION = FirmwareVersion(18, 4, 0)
 
-telemetry_tags_map: dict[str, int] = {
-    "BOARD_ID_HIGH": 1,
-    "BOARD_ID_LOW": 2,
-    "ASIC_ID": 3,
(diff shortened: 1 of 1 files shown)


adjordjevic-TT (approved):

ttexalens/hardware/arc_block.py
```diff
@@ -18,49 +18,8 @@
...
-    "ENABLED_TENSIX_COL": 34,
-    "ENABLED_ETH": 35,
-    "ENABLED_GDDR": 36,
-    "ENABLED_L2CPU": 37,
-    "PCIE_USAGE": 38,
-    "NUMBER_OF_TAGS": 39,
-    "ASIC_LOCATION": 52,
-    "AICLK_LIMIT_MAX": 63,
-}
+# ARC telemetry tags are defined by UMD (tt::umd::TelemetryTag)
```
miacim, line 21: This should not be part of arc block anymore. Quasar do not have arc.

onenezicTT: Aca will follow up with https://github.com/tenstorrent/tt-exalens/issues/1142

PR #1170, opened 2026-09-03
Adding tensix GPRs to Quasar

### LINKED ISSUE #1144: Expose tensix GPRs on Quasar
We should enable dumping tensix GPRs on Quasar. Similar to older architectures we should expose them as registers and create `TensixRegisterDescription` for them. We need to support `neo_id` argument and redefine offset calculation to take into accound `neo_id`.

adjordjevic-TT:
Closes #1144 

Previously we accessed tensix GPR through BRISC because it saw GPRs for all threads. Because of this we calculated offset of `TensixGeneralPurposeRegisterDescription` based on `thread_id`. On Quasar we do not have BRISC so each we have to read tensix GPRs using TRISCs which only see their own thread. To accommodate this `RegisterStore` was changed to grab specific TRISC if thread_id is available instead of default one.

Sadly, on Blackhole there is TRISC2 hardware bug that stops us from reading its private memory properly so I've added workaround for that specific case.

Since now each trisc is responsible for its own thread we do not need to add `thread_id` specific term when calculating register offset so it is removed (except in the workaround).

Added list of tensix GPRs to register map in `functional_neo_block.py`.


ttexalens/register_store.py
```diff
@@ -308,10 +307,25 @@ def read_register(self, register: str | RegisterDescription, safe_mode: bool | N
         elif isinstance(register, ConfigurationRegisterDescription):
             self.location.noc_write32(self._control_register_address, register.index, safe_mode=safe_mode)
             value = self.location.noc_read32(self._data_register_address, safe_mode=safe_mode)
+        elif isinstance(register, TensixGeneralPurposeRegisterDescription):
+            block = self.device.get_block(self.location)
+            assert register.private_address is not None, "Register must have a private address for reading."
+            if self.device.is_blackhole() and register.thread_id == 2:  # Workaround for issue #528
+                risc_debug = block.get_default_risc_debug(
+                    neo_id=self.neo_id
+                )  # We cannot use TRISC2 due to hardware bug so we use BRISC
```
tt-vjovanovic: Don't use `get_default_risc_debug()`, be explicit. If you need brisc, get brisc. Default can be changed for any reason and seeing this bug will be ultra hard.

ttexalens/register_store.py
```diff
@@ -126,8 +126,7 @@ class TensixGeneralPurposeRegisterDescription(RegisterDescription):
     thread_id: int = 0
 
     def __post_init__(self):
-        # Tensix has 64 4 byte general purpose registers per thread.
-        self.offset = self.offset + (self.index + self.thread_id * 64) * 4
```
tt-vjovanovic, line 130: What was original intent for WH/BH for this? Please check if it is safe to change this.

tt-vjovanovic (approved):

ttexalens/register_store.py
adjordjevic-TT: We read them through BRISC and since BRISC sees tensix GPRs of all threads `thread_id` needs to be a part of offset formula. Now each TRISC reads its own thread's tensix GPRs on same memory locations so this `thread_id * 64` part has to be removed.

For the same reason, we are adding that part when we use BRISC instead of TRISC2 on blackhole due to hardware bug.

PR #1171, opened 2026-09-04
Enabling writing to all addresses in trisc2 private memory on blackhole

adjordjevic-TT:
It was confirmed that hardware bug regarding trisc2 private memory access on blackhole is related to read only.
Two tests were conducted:
1. write pattern to trisc2 private memory via NoC and read it through debug hardware
2. write pattern to trisc2 private memory via debug hardware and read it through NoC

First tets failed and second one passed, confirming that only read does not work therefore we are removing assert preventing us from writing to `addresses mod 16 > 4`.


ttexalens/hardware/blackhole/baby_risc_debug.py
```diff
@@ -39,7 +39,6 @@ def read_memory_bytes(self, address: int, buffer: bytearray | memoryview, safe_m
         super().read_memory_bytes(address, buffer, safe_mode=safe_mode)
 
     def write_memory_bytes(self, address: int, data: bytes | bytearray | memoryview, safe_mode: bool | None = None):
-        self.assert_trisc2_address(address)
```
tt-vjovanovic, line 42: Can you reenable or add test for BH write on trisc2 private memory

tt-vjovanovic (approved):

miacim (approved):

PR #1175, opened 2026-09-10
Hidding commands on unsupported archs

adjordjevic-TT:
Adding possibility of architecture specific commands in CLI app. Marking `dump-overlay-state` as only supported for `Quasar`. This means it will not show up when user enters `help`. If user specifically uses command it for unsupported architecture we will print error saying that.

### DIFF
--- a/ttexalens/cli_commands/dump-overlay-state.py
+++ b/ttexalens/cli_commands/dump-overlay-state.py
@@ -38,8 +38,6 @@ Examples:
 
 from typing import Any
 
-import tt_umd
-
 from ttexalens import util
 from ttexalens.context import Context
 from ttexalens.coordinate import OnChipCoordinate
@@ -51,12 +49,19 @@ from ttexalens.rich_formatters import formatter, console
 from ttexalens.hardware.quasar.functional_overlay_registers_description import OverlayRegistersDescription
 from ttexalens.command_parser import CommandMetadata, tt_docopt, CommonCommandOptions
 
+
+def is_supported(device: Device) -> bool:
+    """The functional overlay this command dumps only exists on Quasar."""
+    return device.is_quasar()
+
+
 command_metadata = CommandMetadata(
     short_name="overlay",
     type="low-level",
     description=__doc__,
     common_option_names=[CommonCommandOptions.Device, CommonCommandOptions.Location, CommonCommandOptions.Verbose],
-    supported_archs=[tt_umd.ARCH.QUASAR],
+    is_supported=is_supported,
+    requirement="Quasar devices",
 )
 
 GROUPS = ["counters", "cmdbuf", "errors", "wdt", "debug", "clint", "plic"]
--- a/ttexalens/cli_commands/dump-overlay-state.py
+++ b/ttexalens/cli_commands/dump-overlay-state.py
@@ -50,9 +50,9 @@ from ttexalens.hardware.quasar.functional_overlay_registers_description import O
 from ttexalens.command_parser import CommandMetadata, tt_docopt, CommonCommandOptions
 
 
-def is_supported(device: Device) -> bool:
-    """The functional overlay this command dumps only exists on Quasar."""
-    return device.is_quasar()
+def is_supported(ui_state: UIState) -> bool:
+    """The overlay block this command dumps only exists on Quasar."""
+    return ui_state.current_device.is_quasar()
 
 
 command_metadata = CommandMetadata(
--- a/ttexalens/cli_commands/dump-overlay-state.py
+++ b/ttexalens/cli_commands/dump-overlay-state.py
@@ -52,7 +52,7 @@ from ttexalens.command_parser import CommandMetadata, tt_docopt, CommonCommandOp
 
 def is_supported(ui_state: UIState) -> bool:
     """The overlay block this command dumps only exists on Quasar."""
-    return ui_state.current_device.is_quasar()
+    return hasattr(ui_state.current_block, "overlay")
 
 
 command_metadata = CommandMetadata(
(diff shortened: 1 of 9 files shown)


onenezicTT (approved):

ttexalens/cli_commands/dump-overlay-state.py
```diff
@@ -55,6 +56,7 @@
     type="low-level",
     description=__doc__,
     common_option_names=[CommonCommandOptions.Device, CommonCommandOptions.Location, CommonCommandOptions.Verbose],
+    supported_archs=[tt_umd.ARCH.QUASAR],
```
tt-vjovanovic: I'm wondering if we should replace `supported_archs` with field that will have lambda to check if command is supported on specified device. Currently not all scenarios support different architectures across devices (like Metal), but that doesn't mean tt-exalens should be similar. What I'm aiming at is that we can have scripts that don't support something specific (like P100 that is BH, but doesn't have ETH cores). Then instead of having all those fields that should be checked generically, we can have callback for scripts that will check if selected device is supported.
Also, I'm not sure if we should hide scripts, or just tag them as unsupported on selected device.

ttexalens/cli_commands/dump-overlay-state.py
```diff
@@ -50,11 +49,19 @@
 from ttexalens.hardware.quasar.functional_overlay_registers_description import OverlayRegistersDescription
 from ttexalens.command_parser import CommandMetadata, tt_docopt, CommonCommandOptions
 
+
+def is_supported(ui_state: UIState) -> bool:
+    """The overlay block this command dumps only exists on functional workers on Quasar."""
+    return hasattr(ui_state.current_block, "overlay")
+
+
```
tt-vjovanovic, line 57: If you want to make `is_supports` checking `NocBlock` instead of `Device`, then you should support `--loc` (`-l`) as well. Either way, argument shouldn't be `UIState`. If you think that granularity should be smaller (to include `neo` index), you should add that as an argument.

ttexalens/cli_commands/dump-overlay-state.py
```diff
@@ -50,11 +49,19 @@
...
 command_metadata = CommandMetadata(
     short_name="overlay",
     type="low-level",
     description=__doc__,
     common_option_names=[CommonCommandOptions.Device, CommonCommandOptions.Location, CommonCommandOptions.Verbose],
+    is_supported=is_supported,
+    requirement="Overlay block with Rocket cores",
```
tt-vjovanovic, line 64: I'm not sure if this is useful. When using `--help` you get docstring which will include what is this command used for. That should include text that describes requirement. For example, this command says:
```
Dumps Quasar overlay state at the given location and device....
```
And now in `requirement` you repeat that...

If you think it is important to know which condition wasn't satisfied with `is_supported`, you can rename it to `check_requirements` that should raise exception for every failed check.

adjordjevic-TT: When user types `--help` he will just get message saying command is not supported for current state. Idea behind requirement is, like you said, to specify which condition has not be met so user can better understand why specific command does not work. I will rename it and set it to raise if check fails.

ttexalens/cli_commands/dump-overlay-state.py
adjordjevic-TT: In my mind, granularity should for sure be smaller because that way we allow use more freedom to specify what is required to support a command. Since we are checking against current state `UIState` seemed like logical argument here. The part I don't fully understand is if you are against checking against current state or just against having `UIState` here in the command? We can achieve the same result if here we do device, noc_block, neo, risc ... and then just pass it current state values in `command_parser.py`.

tt-vjovanovic: Imagine this scenario:
> I'm script writer. I wrote method to test if my script is supported for some arguments. Do I need to implement checks along the script as well, or is that `is_supported` method enough?

In other words, since script writer provided `is_supported` method, will infrastructure going to check if it is supported and write message it is not supported, etc. Imagine you call `dump-overlay-state` command with `-loc 0,0 -loc d1,0` where letter `d` was typo. Do you expect nice error message saying script is not supported for the specified location? Do you want script owner to do so, or will infrastructure do it for them by using `is_supported` script?

ttexalens/cli_commands/dump-overlay-state.py
tt-vjovanovic: I was thinking more in a way that `--help` will print everything the same as for when script is working + red text error message saying that it is not currently supported on selected `UIState`.

adjordjevic-TT: That is what currently happens, requirement is only used if user specifically use unsupported command and then we provide more info using requirement.

adjordjevic-TT: Removed requirements.

ttexalens/cli_commands/dump-overlay-state.py
adjordjevic-TT: Now this just checks against current state, checking for arguments remains on script user.

ttexalens/cli_commands/dump-tensix-state.py
```diff
@@ -52,6 +52,12 @@
 )
 from ttexalens.command_parser import CommandMetadata, tt_docopt, CommonCommandOptions
 
+
+def is_supported(device: Device, location: OnChipCoordinate, neo_id: int | None = None) -> bool:
+    print(device.get_block_type(location))
```
tt-vjovanovic: Remove this?

tt-vjovanovic (approved):

PR #1177, opened 2026-09-18
Adding debug bus signals for Quasar

### LINKED ISSUE #968: Add debug bus signals for neo cluster
Currently we only read triscs' pcs through debug bus on quasar. We should add rest of available debug bus signals that can be found [here](https://tenstorrent.atlassian.net/wiki/spaces/TA/pages/696222148/Debug+Daisy+Bus+Stop). Each NEO should have its own debug bus instance.

Signals should be added to `functional_neo_debug_bus_signals.py`.

adjordjevic-TT:
Closes #968 

Adding debug bus signals for Quasar. On quasar debug bus signal map is per NEO instead per block. Added signal groups.
Note that reading signal groups atomically through L1 sampling is not supported so we throw there. There is still value in having signal groups since command use them to group signals in output.

### DIFF
--- a/ttexalens/hardware/quasar/functional_neo_debug_bus_signals.py
+++ b/ttexalens/hardware/quasar/functional_neo_debug_bus_signals.py
@@ -6,8 +6,396 @@ from ttexalens.debug_bus_signal_store import DebugBusSignalDescription
 
 
 debug_bus_signal_map = {
-    "trisc0_pc": DebugBusSignalDescription(rd_sel=1, daisy_sel=7, sig_sel=2 * 5 + 1, mask=0x3FFFFFFF),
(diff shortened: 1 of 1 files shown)


ttexalens/hardware/quasar/functional_neo_debug_bus_signals.py
```diff
@@ -4,10 +4,499 @@
 
 from ttexalens.debug_bus_signal_store import DebugBusSignalDescription
 
-
 debug_bus_signal_map = {
```
tt-vjovanovic, line 7: Current signals seem related only to TRISC0/1/2/3. Are there some Tensix specific signals?

tt-vjovanovic (approved):

ttexalens/hardware/quasar/functional_neo_debug_bus_signals.py
adjordjevic-TT: There are, added those as well.

